PR #44, opened 2012-08-02
Support rebar binary in system

iwamatsu:
This provides the binary of rebar. However, I think that using the rebar binary
installed in the system in many cases.
When rebar is already installed in system, this change that rebar of System use
installed. When not provided, use the binary provide in source.

Signed-off-by: Nobuhiro Iwamatsu iwamatsu@nigauri.org

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,11 @@
 # Author:      Manolis Papadakis
 # Description: Instructions for make
 
+REBAR_BIN = $(shell which rebar)
+ifeq ($(REBAR_BIN),)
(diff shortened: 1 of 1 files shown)


kostis:
After a bit less than 7 years, I finally created a pull request (#211) that does the above for building PropEr but using `rebar3` instead of `rebar`.  This will be merged soon.

Thanks for your contribution!  As they say, "Good things come to those who wait." :-)

kostis:
#211 is now merged into master, to I am closing this.

PR #73, opened 2013-08-06
PropEr passes ultra-paranoid dialyzer checks for R15B02+

istr:
I put some work into adding/updating specs such that PropEr will pass strict compile and dialyzer checking, both for source and eunit tests.
It passes automatic testing with Travis CI for R15B02, R15B03, R16B (and locally R16B01) https://travis-ci.org/istr/proper

Notes:
- the unit tests "run_postcondition_false" and "run_exception" in upstream are never triggered (since the function names do not end with _test)
- the test "rec_props_test1:prop_1" passes locally but fails to pass Travis CI (travis-ci.org). I have no clue why, maybe you could tell?
- some of the tests emit "this clause cannot match because a previous clause at line XXX always matches", I could not yet figure out why, but marked the code TODO there

Please check and merge if you like it. Lots of the latest commits are only there to figure out how to make the build environment play well with Travis.

Please check and pull if you like.

---

side note:
During the heavy testing with dialyzer I found that R15B03 was the last version that did not leak memory. Building a complex .plt with R16B repeatably showed leaky behaviour that froze my machine eventually (due to heavy swapping). It could be alleviated with some  erlang:garbage_collect/0 calls throughout the code, but it really massively "eats" my RAM somewhere.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,4 +13,4 @@ doc/edoc-info
 include/compile_flags.hrl
 .directory
 .eunit
-.dialyzer_plt
+.dialyzer.plt
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 *.beam
 *.dump
 *.tar.gz
+ebin/
 deps/
 doc/*.css
 doc/*.html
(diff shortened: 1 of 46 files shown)


kostis:
Thanks for your PR -- sent long ago.  I think that PropEr's code base is quite OK as far as dialyzer testing is concerned and there is really no need for "ultra-paranoid checks".

PR #74, opened 2013-09-15
Support EEP37: Funs with names

nox:
R17 will [probably](http://erlang.org/pipermail/eeps/2013-September/000466.html) include [my implementation of EEP37](https://github.com/nox/otp/compare/erlang:maint...eep37) which allows expressions of the form:

``` erlang
fact() ->
    fun Fact(N) when N > 0 -> N * Fact(N - 1); Fact(0) -> 1 end
```

PropEr should be able to cope with these expressions. I would like to test this but I don't know which case I should extend.

### DIFF
--- a/src/proper_prop_remover.erl
+++ b/src/proper_prop_remover.erl
@@ -112,6 +112,8 @@ safe_expr({'receive',_Line,Clauses,AfterExpr,AfterBody}) ->
     andalso lists:all(fun safe_expr/1, AfterBody);
 safe_expr({'fun',_Line,{clauses,Clauses}}) ->
     lists:all(fun safe_clause/1, Clauses);
+safe_expr({named_fun,_Line,_Name,Clauses}) ->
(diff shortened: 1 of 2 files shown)


nox:
Oh well.

kostis:
Hi -- coming back to life in PropEr...  was pretty swamped with other stuff till recently.
What is the status of named_funs in OTP these days?  Are they supported starting from 17.0?

If yes, can we have a test added to this pull request (I guess `test/proper_tests.erl` may be an appropriate place to add such a test) and then include it?

nox:
Sorry but ping timeout, I don't really do Erlang anymore.

kostis:
A bit sad, but quite understandable...

Hope you are having fun, whatever you may be doing these days!

nox:
Thanks! I'm having fun working on Servo. Maybe I'll write some Rust-Erlang glue and I'll be able to go to some Erlang conference to present it, one day. :)

PR #90, opened 2014-09-19
Make selftests work with R14B01

hdiedrich:
At the cost of using tc(F, []) instead of tc(F).

R14B01 doesn't even work anymore with the enclosed rebar but it can
be useful for apps using proper that still rely on R14B01, and that
have proper as one of their deps, triggering it to run with `./rebar
eunit`, which otherwise works fine.


kostis:
Thanks for your pull request, but it is not going to be merged. I am now closing it.

PR #97, opened 2015-04-30
Generate compile_flags.hrl also on GNU/Hurd

pinotree:
It is a Unix platform using GNU make, so PropEr can be built there too.


kostis:
Thanks for your contribution

PR #101, opened 2015-07-29
Support 'user_type' in 18.0

garret-smith:
Erlang 18.0 introduced (or corrected, the 17.x docs mention it) the user_type tag in the abstract format.  The structure of type and user_type are the same, so they can be processed in the same way.


kostis:
This was on my TODO list, but I was on vacation and you beat me to it.  Thanks for your contribution!

PR #102, opened 2015-08-24
try to fix non-proper lists in state

matthiaskr:
I encountered the problem a few times that I cannot have non-proper lists in my symbolic state.  As I like to use a dictionary which might include non-proper lists, I made this little fix.  It works for me.  As far as I can tell it doesn't break anything.

### DIFF
--- a/test/proper_tests.erl
+++ b/test/proper_tests.erl
@@ -858,8 +858,8 @@ true_props_test_() ->
      {timeout, 20, ?_passes(symb_statem:prop_parallel_simple())},
      {timeout, 10, ?_passes(ets_statem:prop_ets())},
      {timeout, 20, ?_passes(ets_statem:prop_parallel_ets())},
-     {timeout, 20, ?_passes(pdict_fsm:prop_pdict())}],
-     ?_passes(lists_statem:prop_simple()).
+     {timeout, 20, ?_passes(pdict_fsm:prop_pdict())},
+     ?_passes(lists_statem:prop_simple())].
 
 false_props_test_() ->
     [?_failsWith([[_Same,_Same]],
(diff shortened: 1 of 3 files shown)


kostis:
Thanks for your pull request.  We would appreciate if it also came with some (small) test that showed some case that does not run without this patch but works as expected with it.  This way, the chances of not breaking this feature in the future are maximized.  Looking forward its inclusion.

kostis:
Hello again.

It is possible to submit a test case for this pull request?

matthiaskr:
@kostis sorry that I didn't respond (and forgot about that).  I'll look into adding a test to the pull request.

matthiaskr:
It will take a bit longer as the normal tests won't run smooth on my system (even master branch).  Is there any trouble using it on Erlang 18.3?  I'm just wondering if I should downgrade...

matthiaskr:
@kostis worked around my problem and added a test.

test/proper_tests.erl
```diff
@@ -858,7 +858,8 @@ true_props_test_() ->
      {timeout, 20, ?_passes(symb_statem:prop_parallel_simple())},
      {timeout, 10, ?_passes(ets_statem:prop_ets())},
      {timeout, 20, ?_passes(ets_statem:prop_parallel_ets())},
-     {timeout, 20, ?_passes(pdict_fsm:prop_pdict())}].
+     {timeout, 20, ?_passes(pdict_fsm:prop_pdict())}],
+     ?_passes(lists_statem:prop_simple()).
```
kostis: This test seems a bit wrongly placed.  Notice that all others are inside the list, while this one is outside.  Can you place it before or after the line which reads: `?_passes(pdict_statem:prop_pdict())`  (if you have such a line in your fork)?

kostis:
Thanks for the test.  As far as I know, the tests of PropEr work OK on Erlang 18.3.

See also the build history at: https://travis-ci.org/manopapad/proper

You may want to send me, perhaps off-line, some more info about your system and which tests do not run smoothly there.

Thanks again for your work!

test/proper_tests.erl
matthiaskr: a crap, will change...

PR #108, opened 2015-11-20
Enabling ?LET macros to use user defined types

TheGeorge:
This fixes an issue where ?LET could not use user defined types like this:

    -type my_type() :: integer().
    even_numbers() ->
        ?LET(I, my_type(), I*2).


TheGeorge:
The tests 'fail' (unexpected success) because of the added functionality. Should I fix the tests?

StoneCypher:
:+1: +1

PR #119, opened 2016-04-03
maps: support in symbolic vars

fenollp:
Fix for part 3 of https://github.com/manopapad/proper/issues/118

Any plan on supporting OTP > 17?

I encountered the issue solved when writing a proper_statem that get/set'ed maps.
More precisely, `next_state/3`'s `V` was applied `{call,...}`s to extract values that were then stored in a map (then stored in a `state` record).

However in subsequent runs of commands, the symbolic calls would never be turned into "dynamic values". Errors such as `badarg: ++/2 can't work on {call,maps,get,[k,#{k=>v}]} []`

This does not solve the whole "support maps" issue, nor is the fastest implementation.

### DIFF
--- a/src/proper_symb.erl
+++ b/src/proper_symb.erl
@@ -295,13 +295,29 @@ symb_walk_call(VarValues, Mod, Fun, Args,
     HandleCall(Mod, Fun, HandledArgs).
 
 -spec symb_walk_gen(var_values(), symb_term(), handle_info()) -> handled_term().
+-ifdef(AT_LEAST_17).
 symb_walk_gen(VarValues, SymbTerm,
-	      {_Caller,_HandleCall,HandleTerm} = HandleInfo) ->
+              {_Caller,_HandleCall,HandleTerm} = HandleInfo) ->
     SymbWalk = fun(X) -> symb_walk(VarValues, X, HandleInfo) end,
     Term =
-	if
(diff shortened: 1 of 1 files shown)


fenollp:
Added compile-time check that defines macro `maps_support` iff the running OTP supports maps

fenollp:
Ah I only saw now the "non standard" `write_compile_flags` script. Maybe I should revert to use that one?

fenollp:
@kostis @manopapad  Ping

kostis:
There are various issues here.  For starters, the latest `master' branch does support 17.x and 18.x.

It's true that PropEr does not support maps.  Supporting maps fully is not trivial and it's also a bit of a moving target since the type language of maps has been extended/expanded (and improved!) in 19.  Unfortunately, supporting maps in different releases also means that the `term()' type needs to mean different things in different releases.  We simply do not have the manpower to do something like that.
If i had to choose a type language to support it would be that of 19 and my current plan is to tag a 1.2 (1.x) versions of PropEr that will support OTP releases prior to 19 and start a new 2.x line that will only support 19.x and higher.

Now, coming back to this pull request, I do not see anything fundamentally wrong with it (other than the code duplication) and, provided it gets enhanced with test(s) that show its usefulness and ensure that this feature (supporting maps in symbolic vars) continues to work after changes, I could go ahead and merge it.  But I am reluctant to accept pull requests for which we have no tests.

PR #125, opened 2016-06-14
Bump proper.app.src to 1.2 to match the new tag

jaynel:
You will need to move the tag forward one commit.


kostis:
Thanks @jaynel !

Issue #118, opened 2016-04-03
Internal support for maps


fenollp:
Maps are still not supported, be it
1. in the typeserver https://github.com/manopapad/proper/pull/96
1. in the generator https://github.com/manopapad/proper/issues/111
1. in [symb_walk_gen](https://github.com/manopapad/proper/blob/1b773eeb47cb2c3116d78bdf681505703b762eee/src/proper_symb.erl#L303-L305)

I am posting my solution to the last point after this.

PR #128, opened 2016-06-25
Make next_state/3 handle maps

fenollp:
Rewrite of #119 & adding a test.

Note: this is not "whole maps support" as #119's top line hinted: only a fix for part 3 of #118.

The test is a copy of symb_statem that reads from a map in a symbolic value and updates the SUT's map state.

### DIFF
--- a/src/proper_symb.erl
+++ b/src/proper_symb.erl
@@ -299,10 +299,26 @@ symb_walk_call(VarValues, Mod, Fun, Args,
 symb_walk_gen(VarValues, SymbTerm,
 	      {_Caller,_HandleCall,HandleTerm} = HandleInfo) ->
     SymbWalk = fun(X) -> symb_walk(VarValues, X, HandleInfo) end,
-    Term =
-	if
-	    is_list(SymbTerm)  -> proper_arith:safe_map(SymbWalk, SymbTerm);
-	    is_tuple(SymbTerm) -> proper_arith:tuple_map(SymbWalk, SymbTerm);
-	    true               -> SymbTerm
-	end,
+    Term = do_symb_walk_gen(SymbWalk, SymbTerm),
     HandleTerm(Term).
+
+-spec do_symb_walk_gen(fun((T) -> S), maybe_improper_list(T,T | [])) ->
+			  maybe_improper_list(S,S | []).
+-ifdef(AT_LEAST_17).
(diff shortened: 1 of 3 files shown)


kostis:
Some comments:

- The style of the code is TERRIBLE.  Yes, I know this is subjective.  The objective comment is that you have not followed at all the style that the rest of the code uses.  Please do so.

 - This is the first time ever that a contributor adds his name to the THANKS file.  You can NOT do that.  Only _we_ modify this file!

fenollp:
* Could you point to the impeding coding style document / particular issues?
* I assumed that was how you guys handled it. No need for passive aggressiveness. I did not intend to hurt feelings.
* Right. I was a bit lost with all the names. Amending now.

kostis:
- Take a look at [proper_tests.erl](https://github.com/manopapad/proper/blob/master/test/proper_tests.erl). For starters, there is no single line, I think, that has `,` in the beginning of the next line.  Similar for the export list of [symb_statem.erl](https://github.com/manopapad/proper/blob/master/test/symb_statem.erl) from which `symb_statem_maps.erl` is derived.  There is also no line that contains only `]).` or `}).`.  These always exists in the previous line.  Also, the code does not use spaces; it is indented with tabs using emacs.  (There may be other issues...)

 - There are no hurt feelings. It's just that I actually find it weird for somebody to thank himself...

 - Good.

fenollp:
* done
* haha
* :)

kostis:
Do two more things and then I can merge this:
 1. What I meant with "indent with tabs using emacs" does not mean literally that the file should contain tabs in each line.  It's what you get if you press tab, while in emacs, on each line if you have the Erlang mode enabled.  (You also get a similar outcome if you edit the file in emacs and do ^X indent-buffer).  If it is too complicated, leave it as is and I can fix that myself.
 2. Please squash the commits into one.  (I.e. `git fetch; git rebase -i origin/master` and squash commits)

fenollp:
1. I am using emacs and erlang-mode. I tried disabling my turn-tabs-to-spaces code but that didn't work, so I tried doing it by hand…
2. done!

kostis:
Fixed indentation with a new commit (e90343e).

fenollp:
Yay thanks! :)

PR #136, opened 2017-01-30
fix rebar.config and clean_temp.sh to work properly on solaris

ztmr:


kostis:
Thanks for your contribution. Unfortunately, we do not test Solaris these days...

PR #138, opened 2017-02-09
Implementation of a ?SETUP macro

### LINKED ISSUE #115: ?SETUP macro
It would be great to have a ?SETUP macro to wrap properties (similar to QuickCheck) so that you can setup and cleanup the full run (and not each single test) "PropErly"

TheGeorge:
This implements a ?SETUP macro that works as follows:
```erlang
?SETUP(fun () ->       % setup function
           do_some_setting_up(),
           fun () ->   % finalize function
               ok = do_some_finalizing(),
           end
       end,
       ?FORALL( ... ))
```

The setup function is called an is expected to return a finalize function. The setup function is called before the first test to setup a test environment. The finalize function is then called after the last test.

This also fixes #115

### DIFF
--- a/src/proper.erl
+++ b/src/proper.erl
@@ -674,9 +674,9 @@ global_state_erase() ->
 setup_test(#opts{setup_funs = Funs}) ->
 	[Fun() || Fun <- Funs].
 
--spec finalize_test([finalize_fun()]) -> ['ok'].
+-spec finalize_test([finalize_fun()]) -> 'ok'.
 finalize_test(Finalizers) ->
-	[ok = Fun() || Fun <- Finalizers].
+	lists:foreach(fun (Fun) -> ok = Fun() end, Finalizers).
 
 %% @private
 -spec spawn_link_migrate(fun(() -> 'ok')) -> pid().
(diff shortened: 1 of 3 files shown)


test/proper_tests.erl
```diff
@@ -1039,6 +1039,57 @@ options_test_() ->
 		 ?FORALL(_,?SIZED(Size,integer(Size,Size)),false),
 		 [{start_size,12}])].
 
+setup_prop() ->
+  ?SETUP(fun () -> put(setup_token, true),
```
kostis: Minor stylistic comment: Do not write
```erlang
  fun () -> some_code(),
            more_code()
  end
```
Instead write:
```erlang
  fun () ->
      some_code(),
      more_code()
  end
```

src/proper.erl
```diff
@@ -358,10 +363,10 @@
 -export([get_size/1, global_state_init_size/1,
 	 global_state_init_size_seed/2, report_error/2]).
 -export([pure_check/1, pure_check/2]).
--export([forall/2, implies/2, whenfail/2, trapexit/1, timeout/2]).
+-export([forall/2, implies/2, whenfail/2, trapexit/1, timeout/2, setup/2]).
 
 -export_type([test/0, outer_test/0, counterexample/0, exception/0,
-	      false_positive_mfas/0]).
+	      false_positive_mfas/0, setup_fun/0, finalize_fun/0]).
```
kostis: It's not so clear why these types are exported; they do not seem to be used (or documented) anywhere...

TheGeorge:
I fixed the code according to your comments. By the way is there a reason why both tabs and spaces are used instead of just one of them? (see function get_size/1 for example)

PR #141, opened 2017-02-22
Increase adts_test's timeout for testing on slow arches

### LINKED ISSUE #139: adts_test_ in test/proper_tests.erl times out on slower arches
Hello!

I am the package maintainer for proper in Fedora, and I've found that ```adts_test_``` in ```test/proper_tests.erl``` will fail on two arches: ```armv7hl``` and ```aarch64``` due to the test timing out. The test runner will end like this:

```
    proper_tests:1020: options_test_...[0.005 s] ok
    proper_tests:1023: options_test_...[0.005 s] ok
    proper_tests:1026: options_test_...[0.001 s] ok
    proper_tests:1027: options_test_...[0.001 s] ok
    proper_tests:1028: options_test_...[0.001 s] ok
    proper_tests:1029: options_test_...ok
    proper_tests:1032: options_test_...[0.001 s] ok
    proper_tests:1046: adts_test_...[0.250 s] ok
    proper_tests:1049: adts_test_...*timed out*
undefined
=======================================================
  Failed: 0.  Skipped: 0.  Passed: 447.
One or more tests were cancelled.
DEBUG: Reconstruct stdlib []
DEBUG: Reconstruct kernel [{error_logger,tty}]
DEBUG: Reconstruct rebar [{log_level,warn},
                          {log_colored,uncolored},
                          {any_dir_modules,
                              [rebar_require_vsn,rebar_deps,rebar_subdirs,
                               rebar_templater,rebar_cleaner]},
                          {modules,
                              [{app_dir,
                                   [rebar_proto_compiler,
                                    rebar_protobuffs_compiler,
                                    rebar_neotoma_compiler,
                                    rebar_asn1_compiler,rebar_dia_compiler,
                                    rebar_dialyzer,rebar_erlc_compiler,
                                    rebar_lfe_compiler,rebar_erlydtl_compiler,
                                    rebar_port_compiler,rebar_otp_app,
                                    rebar_otp_appup,rebar_ct,rebar_eunit,
                                    rebar_qc,rebar_escripter,rebar_edoc,
                                    rebar_shell,rebar_xref,rebar_metacmds]},
                               {rel_dir,
                                   [rebar_appups,rebar_reltool,
                                    rebar_upgrade]}]},
                          {recursive_cmds,
                              ['check-deps',compile,'delete-deps','get-deps',
                               'list-deps','update-deps']}]
ERROR: One or more eunit tests failed.
ERROR: eunit failed while processing /builddir/build/BUILD/proper-1.2: rebar_abort
```

I've worked around this by renaming the test to adts_dontrun_ with sed when the test suite is run on these arches, but if there's a way for us to extend the timeout so the test can complete that would be helpful. Thanks!

kostis:
An attempt to fix #139 -- let's see if the increase is sufficient.


bowlofeggs:
With this patch I am able to build proper on all arches, thanks!

bowlofeggs:
Oh no, I lied to you. I had removed the "if arch" statements but had forgotten to remove my sed statement that patched out the test. So my builds were done with that test skipped on all arches and were wrong. When I realized my mistake I retested and the test still fails. So hold on a bit and I'll fiddle with the numbers until they work and I'll tell you what I find.

bowlofeggs:
I've tried setting these values to pretty large numbers and they seem to fail. I'm starting to wonder if the timeout is from something else. For example, I set it to 4096 and 8192 for this build:

https://kojipkgs.fedoraproject.org//work/tasks/5792/18005792/build.log

bowlofeggs:
I assume the unit of the timeout is in seconds? If that is so, I believe something else is causing the timeout than these values as the job that this log came from ran for about 5 minutes (far less than 4096 seconds). You can see the start and end times here:

https://koji.fedoraproject.org/koji/taskinfo?taskID=18005792

kostis:
The timeout unit is in milliseconds. But these tests do not take that long to run, at least on the machines I have access to, so there may be something else which causes this.

Looking at the build log that you've sent I see that it's pretty complicated.  PropEr only needs a simple 'make all' command to build and test everything and also comes wiith its own copy of the `rebar` script, while you seem to be using the globally installed one:
<pre>
...
+ /usr/bin/rebar compile skip_deps=true -vv
DEBUG: Consult config file "/builddir/build/BUILD/proper-1.2/rebar.config"
DEBUG: Rebar location: "/usr/bin/rebar"
DEBUG: Consult config file "/builddir/build/BUILD/proper-1.2/src/proper.app.src"
</pre>
Can that be the issue?  What happens if you just git clone the branch of this pull request (https://github.com/manopapad/proper/pull/141) or download the 1.2 tag of proper and try to just run `make all` once you have applied an increase in the timeout values of this test?

test/proper_tests.erl
```diff
@@ -1107,16 +1107,15 @@ setup_test_() ->
...
-     {timeout, 84,	% for 18.x (and onwards?) on slow machines
-      ?_passes(?FORALL({X,Y,D},
-		       {integer(),float(),?DICT(integer(),float())},
+     {timeout, 40,	% for 18.x (and onwards?)
+      ?_passes(?FORALL({X,Y,D}, {integer(),float(),?DICT(integer(),float())},
 		       dict:fetch(X,dict:store(X,Y,eval(D))) =:= Y), [30])},
-     ?_fails(?FORALL({X,D},
-	     {boolean(),?DICT(boolean(),integer())},
-	     dict:erase(X, dict:store(X,42,D)) =:= D))].
+     {timeout, 20,      % seems to be needed on slow machines (ARM)
```
bowlofeggs: I found that this does pass on our aarch64 builder, but still fails on our armv7hl builder. If I change this 20 to 256 it passes on all Fedora architectures. I also tried 128 and that failed on armv7hl as well. So somewhere between 128 and 256 is probably the sweet spot, but if you think 256 is an ok value here I'm happy to just go with that.

bowlofeggs: Well it turns out that not all our arm builders are the same speed. I found that even 512 doesn't always pass. I'm fiddling with the numbers a bit more to see if I can find one that passes more reliably.

bowlofeggs:
I was able to get the build to work on all Fedora architectures with this patch:

http://pkgs.fedoraproject.org/cgit/rpms/erlang-proper.git/tree/0000-Add-timeout-values-that-work-on-ARM.patch?id=eef561e53360eb341528c9f33dc0afaaa0d47932

This is the build that succeeded with that patch:

https://koji.fedoraproject.org/koji/taskinfo?taskID=18058677

kostis:
Seems that all the changes of this PR are already in PropEr's code base.  Closing this without merge.
#139 is fixed.

PR #150, opened 2017-08-03
make term colors optional

cjkjellander:
Add a 'nocolors' option to suppress all term color output for
later parsing by scripts.

We started hitting this in jenkins that chocked on \033 that wasn't a valid utf-8 character.

### DIFF
--- a/src/proper.erl
+++ b/src/proper.erl
@@ -262,6 +262,8 @@
 %%% from the MFA or an exception with it's reason.  If needed, the
 %%% user supplied function can call erlang:get_stacktrace/0.  Default
 %%% is undefined.</dd>
+%%% <dt>`nocolors'</dt>
+%%% <dd>Don't use term colors in output.</dd>
 %%% </dl>
 %%%
 %%% == Spec testing ==
@@ -486,7 +488,8 @@
 		  | 'any_to_integer'
 		  | {'spec_timeout',timeout()}
 		  | {'skip_mfas',[mfa()]}
-		  | {'false_positive_mfas',false_positive_mfas()}.
+		  | {'false_positive_mfas',false_positive_mfas()}
+                  | 'nocolors'.
 
 -type user_opts() :: [user_opt()] | user_opt().
 -record(opts, {output_fun       = fun io:format/2 :: output_fun(),
@@ -503,7 +506,8 @@
 	       spec_timeout     = infinity        :: timeout(),
 	       skip_mfas        = []              :: [mfa()],
 	       false_positive_mfas                :: false_positive_mfas(),
-	       setup_funs       = []              :: [setup_fun()]}).
+	       setup_funs       = []              :: [setup_fun()],
+               nocolors         = false           :: boolean()}).
 -type opts() :: #opts{}.
 -record(ctx, {mode     = new :: 'new' | 'try_shrunk' | 'try_cexm',
 	      bound    = []  :: imm_testcase() | counterexample(),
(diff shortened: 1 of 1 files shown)


src/proper.erl
```diff
@@ -1729,24 +1741,34 @@ apply_skip(Args, Prop) ->
...
-		  #opts{expect_fail = ExpectF, output_fun = Print}) ->
-    case ExpectF of
-	true ->
-	    Print("\033[1;31mFailed: All tests passed when a failure was expected.\033[0m~n", []);
-	false ->
-	    Print("\033[1;32mOK: Passed ~b test(s).~n\033[0m", [Performed])
+                  #opts{expect_fail = ExpectF, output_fun = Print,
+                        nocolors = NoColors}) ->
+    case {ExpectF, NoColors} of
+        {true, true} ->
```
kostis: I suggest that you use two nested `case`s here rather than wrapping `ExpectF` and `NoColors` to a tuple and having a `case` with four branches.

src/proper.erl
```diff
@@ -1729,24 +1741,34 @@ apply_skip(Args, Prop) ->
...
 			performed = Performed},
-		  #opts{expect_fail = ExpectF, output_fun = Print}) ->
-    case ExpectF of
-	true ->
-	    Print("\033[1;32mOK: Failed as expected, after ~b test(s).~n\033[0m", [Performed]);
-	false ->
-	    Print("\033[1;31mFailed: After ~b test(s).~n\033[0m", [Performed])
+                  #opts{expect_fail = ExpectF, output_fun = Print,
+                        nocolors = NoColors}) ->
+    case {ExpectF, NoColors} of
```
kostis: Same comment as above here.

src/proper.erl
```diff
@@ -1756,20 +1778,26 @@ report_imm_result({error,Reason}, #opts{output_fun = Print}) ->
 
 -spec report_rerun_result(run_result(), opts()) -> 'ok'.
 report_rerun_result(#pass{reason = Reason},
-		    #opts{expect_fail = ExpectF, output_fun = Print}) ->
-    case ExpectF of
-	true  -> Print("\033[1;31mFailed: \033[0m", []);
-	false -> Print("\033[1;32mOK: \033[0m", [])
+                    #opts{expect_fail = ExpectF, output_fun = Print,
+                        nocolors = NoColors}) ->
+    case {ExpectF, NoColors} of
```
kostis: Same comment as above here.

src/proper.erl
```diff
@@ -1756,20 +1778,26 @@ report_imm_result({error,Reason}, #opts{output_fun = Print}) ->
...
+        {true, true}  -> Print("Failed: ", []);
+        {true, false}  -> Print("\033[1;31mFailed: \033[0m", []);
+        {false, true} -> Print("OK: ", []);
+        {false, false} -> Print("\033[1;32mOK: \033[0m", [])
     end,
     case Reason of
-	true_prop   -> Print("The input passed the test.~n", []);
```
kostis: We would rather not have whitespace-only changes.

src/proper.erl
```diff
@@ -1756,20 +1778,26 @@ report_imm_result({error,Reason}, #opts{output_fun = Print}) ->
...
 	didnt_crash -> Print("The input didn't raise an early exception.~n", [])
     end;
 report_rerun_result(#fail{reason = Reason, actions = Actions},
-		    #opts{expect_fail = ExpectF, output_fun = Print}) ->
-    case ExpectF of
-	true  -> Print("\033[1;32mOK: \033[0m", []);
-	false -> Print("\033[1;31mFailed: \033[0m", [])
+                    #opts{expect_fail = ExpectF, output_fun = Print,
+                          nocolors = NoColors}) ->
+    case {ExpectF, NoColors} of
```
kostis: Same comment as the use of two nested `case`s here.

cjkjellander:
Review comments fixed.

PR #152, opened 2017-11-18
Support a map generator

ferd:
Simply parametrize it from using two generators for keys and values.
The map is built from using a list of key/value pairs and converting
them.

This implementation was chosen for its simplicity, but also because it
allows to use a PropEr version that supports maps in an older Erlang
version without compile errors (nor runtime errors as long as the
generator is not used).

### DIFF
--- a/src/proper_types.erl
+++ b/src/proper_types.erl
@@ -139,7 +139,7 @@
 
 -export([integer/2, float/2, atom/0, binary/0, binary/1, bitstring/0,
 	 bitstring/1, list/1, vector/2, union/1, weighted_union/1, tuple/1,
-	 loose_tuple/1, exactly/1, fixed_list/1, function/2, any/0,
+	 loose_tuple/1, exactly/1, fixed_list/1, function/2, map/2, any/0,
 	 shrink_list/1, safe_union/1, safe_weighted_union/1]).
 -export([integer/0, non_neg_integer/0, pos_integer/0, neg_integer/0, range/2,
 	 float/0, non_neg_float/0, number/0, boolean/0, byte/0, char/0,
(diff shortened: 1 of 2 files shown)


kostis:
Thanks for your pull request.  It's certainly a (good) start, one that should have been done long ago, and perhaps at that point it would have sufficed. Today, it does not. One of the salient features of PropEr is its tight integration with the type language of Erlang -- in fact this is reflected in [the title of the paper that introduced PropEr](https://dl.acm.org/citation.cfm?id=2034663). What this means in practice, is that in PropEr it is possible for the system to take any type and automagically produce a generator (and a shrinker) for it.

These days the type language of maps has been extended and it's possible to specify more restricted map types than just `map()`.  So, if PropEr is to get (or at least claim that it has) support for maps, IMO it should support the complete subset of the type language that allows specification of more restricted maps.

On the other hand, there is no harm done in including this in the code base.  But under no circumstances should there be a claim/mention anywhere that PropEr supports maps just because of this.  It does not.

src/proper_types.erl
```diff
@@ -1096,6 +1096,15 @@ function(Arity, RawRetType) when is_integer(Arity), Arity >= 0, Arity =< 255 ->
 function(RawArgTypes, RawRetType) ->
     function(length(RawArgTypes), RawRetType).
 
+%% @doc A map whose keys are defined by the generator `K' and values
+%% by the generator `V'.
+-spec map(K, V) -> proper_types:type() when
```
kostis: The style of this spec unnecessarily differs from those used in the file.  Please change it to a one-liner for uniformity.

ferd: Fixed it.

kostis:
I'll go ahead and merge this, since it's a clear improvement.  But, as I wrote, this is only the start for PropEr support for maps.

PR #155, opened 2018-02-17
report-shrinking: report exceptions when shrinking ?FORALLs

fenollp:
A call inside my property was failing (only) during shrinking but its failing was never reported. This makes PropEr io:format/2 some information.

To reproduce:

```erlang
prop_throws_when_shrinking() ->
    ?FORALL(X, ?SHRINK(a, [b])
           ,case X of
                a -> false;
                b -> throw(this_wont_show_without_this_patch)
            end
           ).
```


fenollp:
Tested on 20.2 + rebar3, passes 548 tests although rebar2 passes 556...
I believe Travis just needs a larger timeout on the heavier tests?

kostis:
I am trying to understand the supplied example (perhaps its phrasing is a bit weird, or it's too just simplistic?) and, more importantly, what is the PropEr(TM) thing to happen here.

- First of all, what does it mean "A call inside my property was failing (only) during shrinking"? Shrinking happens when a property fails for some other reason, right?

- Second, if thrown exceptions during shrinking are reported (and the stack trace gets printed, in the same way that stack traces get printed when the evaluation of a property encounters an exception -- this part at least seems consistent) shouldn't this be treated as a failure and the shrinking should be considered as successful?  I.e., shouldn't the above code return `b` instead of `a` (that it currently does, even with this patch)?

Finally, a detail: can you please rebase this?

fenollp:
Ah sorry about that! So the idea of using `?SHRINK(Gen, [Shrinker])` is to
1. generate from Gen as long as the property passes
1. generate from Shrinker as soon as the property fails once
So what happens here is that as soon as `a` is generated, the property is made to fail. Then PropEr shrinks using the second generator, thus generating b which is made to throw inside the property. 
Indeed one could see that throw as a failure of the property and I hadn’t thought of that.
My idea of a failing property was of one that returns false instead of true, leaving exceptions as neither these interpretation but instead as errors in the testing code.

What do you think of adding an opt named something along “only_boolean_properties” which, if enabled on a FORALL, would print exceptions with this patch then stop, thus no longer considering exceptions as failures of the property?

I have the worst network right now, will rebase as soon as the cable gets dry!

kostis:
> Ah sorry about that! So the idea of using ?SHRINK(Gen, [Shrinker]) is to
>  1. generate from Gen as long as the property passes
>  2. generate from Shrinker as soon as the property fails once

Right.  Here is a property that shows this:
```erlang
prop_int_fails() ->
  ?FORALL(N, ?SHRINK(integer(), [integer(-69,-54)]), N > -42).
```
```erl
2> proper:quickcheck(fenoll:prop_int_fails()).
.................................................!
Failed: After 50 test(s).
-47

Shrinking ..(2 time(s))
-54
```
The `integer()` generator is used until one of its instances fails, and then the (presumably _simpler_) generator `integer(-69,-54)` is used for shrinking.  This is the purpose (and intended use) of `?SHRINK`.

> So what happens here is that as soon as a is generated, the property is made to fail. Then PropEr shrinks using the second generator, thus generating b which is made to throw inside the property.
Indeed one could see that throw as a failure of the property and I hadn’t thought of that.
My idea of a failing property was of one that returns false instead of true, leaving exceptions as neither these interpretation but instead as errors in the testing code.

It's not a bad idea, but it's not how it currently works.  Consider:
```erlang
prop_int_throws() ->
  ?FORALL(N, integer(),
          case N of
            _ when N > -17 -> true;
                % crash for all values in between
            _ when N < -42 -> false
          end).
```
```erl
3> proper:quickcheck(fenoll:prop_int_throws()).
................................................!
Failed: After 49 test(s).
An exception was raised: error:{case_clause,-25}.
Stacktrace: [{fenoll,'-prop_int_throws/0-fun-0-',1,
                     [{file,"fenoll.erl"},{line,16}]}].
-25

Shrinking .(1 time(s))
-17
false
```

> What do you think of adding an opt named something along “only_boolean_properties” which, if enabled on a FORALL, would print exceptions with this patch then stop, thus no longer considering exceptions as failures of the property?

I need to think more about this. As I wrote, it's not a bad idea, but I am a bit skeptical about its code complexity vs. real usefulness ratio.  If you have some time to spare, perhaps a pull request (with tests) that shows the amount of added code needed and all places where this option needs to be checked will convince me more easily.

fenollp:
True, that is indeed a simpler example and I needn’t make mine so convoluted. 

Yes I’ll look into doing that when I have some time. I’d also like to pass output_fun and prefix around, maybe in the pdict if needed. 
Thanks for your inputs. 

Oh and yes when using `?SHRINK/2` the list on the right is of generators that are *presumably simpler* than the first one. Is that enforced somehow somewhere? I couldn’t find anything that looked like it compared generators... From what I can tell PropEr just tries them one by one until it gives up, no?

kostis:
> Oh and yes when using ?SHRINK/2 the list on the right is of generators that are presumably simpler than the first one. Is that enforced somehow somewhere? I couldn’t find anything that looked like it compared generators... 

No.  This is not enforced.  First of all, there is no built-in notion of what is a _simpler_ generator (as a matter of fact, the notion is arguably not particularly well-defined; see my example above: why is `integer(L, H)` simpler than `integer()`, just because its range is smaller?).  Second, the purpose of the tool is to test programs, not test how well programmers write properties or how they possibly make mistakes in this task.

> From what I can tell PropEr just tries them one by one until it gives up, no?

Right.

kostis:
@fenollp: This has been open for quite some time. I think I will close this PR without merging it.  The main reason is that, as I wrote in my first message, your example code should really return `b` instead of `a`, i.e., the shrinking needs to be considered as successful in this case.

Having written that, I think that the whole issue is based on an initial misunderstanding (from your part) of what `?SHRINK(..)` is supposed to be doing.  I hope it's more clear now.

PR #156, opened 2018-03-02
setup: only create map once

fenollp:


src/proper.erl
```diff
@@ -715,14 +715,15 @@ setup_test(#opts{output_fun = OutputFun,
 		 start_size = StartSize,
 		 max_size = MaxSize,
 		 setup_funs = Funs}) ->
+    SetupOpts = #{numtests => NumTests,
+				  search_steps => SearchSteps,
```
kostis: The indentation here does not seem right, does it?

fenollp: I tried my best to replicate the style of mixed tabs/spaces of the project & that may have failed. Can also be the way Github displays code though.

kostis: OK, I will merge this, see how it looks on my emacs, and possibly fix it if it's not PropEr.

Issue #157, opened 2018-03-15
Friendlier search macros?


ferd:
Guided by Kostis' keynote this morning at BEAMCode SF, I was reading from the source and some docs, and noticed that:

a) the first method suggested in the talk:

```erlang
?FORALL_SA(GVar, #{gen => MyGen},
    begin
        some_calc(GVar),
        ?MAXIMIZE(GVar),
        my_prop(GVar)
    end).
```
is currently deprecated according to code comments.

b) The alternative format mentioned is instead:
```erlang
?EXISTS(GVar, #{gen => MyGen},
    begin
        some_calc(GVar),
        ?MAXIMIZE(GVar),
        my_prop(GVar)
    end).
```
With a similar variation existing for `?NOT_EXISTS`, but suddenly nothing seems to be there for `?FORALL` that is not deprecated.

I'm curious what the intended usage long term should be. Is there a reason for abandoning the ?FORALL approach and aiming towards `EXISTS` or `NOT_EXISTS`?

 For me, and in a way of making the friendliest thing possible, I was thinking of a macro a bit like:

```erlang
?TARGET_ALL(GVar, RegularGenerator, % no #{gen => G} deal here
    begin
        ?MAXIMIZE(whatever(GVar)),
        my_prop(GVar)
    end).
```
I can imagine easily getting something similar to that with `NOT_EXISTS(GVar, #{gen => G}, not Prop)` but I'm a bit puzzled by the rationale behind the current approach.

TheGeorge:
Hi,
thank you for your input. I can give some insight into why the macros are as they are:

a) TPBT was originally developed as an extension project to PropEr, something that you could plugin when you needed it but also something that wouldn't change how core PropEr is implemented. Like an Addon library. `?FORALL_SA` is actually a short form for:
```
?STRATEGY(simulated_annealing,
          ?FORALL(...)).
```
The `?TARGET` macro hands the generator (and possibly a neighborhood function) to the library, which performs some calculations, and returns a PropEr generator. This generator is then, of course, hooked up to the search strategy.

However, as you mentioned, this is a bit ugly and cumbersome. You need to know how the system works and the macros are not self-explanatory. 

b) The `?EXISTS` and `?NOT_EXISTS` macros where chosen because they allow the same expressiveness as the `?FORALL` macro and they are visually different from them:
 + `?NOT_EXISTS` is the targeted version of `?FORALL`, one just needs to invert the condition of the property.
[...]

PR #158, opened 2018-04-09
search macro improvements (see issue #157)

TheGeorge:
`?EXIST` and `?NOT_EXIST` do not require the `#{gen=>...}` thingy anymore and there is a `?FORALL_TARGETED` that works like the exists and is not specialized on a search strategy (which should be specified in the proper options)

### DIFF
--- a/src/proper_statem.erl
+++ b/src/proper_statem.erl
@@ -218,7 +218,7 @@
 
 -module(proper_statem).
 
--export([commands/1, commands/2, commands/4, parallel_commands/1, parallel_commands/2,
+-export([commands/1, commands/2, parallel_commands/1, parallel_commands/2,
 	 more_commands/2]).
 -export([run_commands/2, run_commands/3, run_parallel_commands/2,
 	 run_parallel_commands/3]).
--- a/src/proper_statem.erl
+++ b/src/proper_statem.erl
@@ -218,7 +218,7 @@
 
 -module(proper_statem).
 
--export([commands/1, commands/2, parallel_commands/1, parallel_commands/2,
+-export([commands/1, commands/2, commands/4, parallel_commands/1, parallel_commands/2,
 	 more_commands/2]).
 -export([run_commands/2, run_commands/3, run_parallel_commands/2,
 	 run_parallel_commands/3]).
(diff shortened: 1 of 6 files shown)


src/proper_statem.erl
```diff
@@ -218,7 +218,7 @@
 
 -module(proper_statem).
 
--export([commands/1, commands/2, parallel_commands/1, parallel_commands/2,
+-export([commands/1, commands/2, commands/4, parallel_commands/1, parallel_commands/2,
```
fenollp: Where is this used by the "targeted" part of PropEr?

TheGeorge: I should read my diffs more carefully (or maybe don't make pull requests that late). Thanks for pointing it out.

fenollp:
Note: Travis CI's `make all` probably needs a `rebar3 xref` equivalent. Would have caught that typo :)

TheGeorge:
So after adding the test case, I discovered a bug, where shrinking did not work, when TPBT was used together with a constructed '?LET' generator. This PR should fix this bug (I also added a test to check for this).

TheGeorge:
I also played around with the test macros and they should now not timeout anymore.

kostis:
> I also played around with the test macros and they should now not timeout anymore.

This is nice, of course, but do we have any idea why the timeout was happening?  Using 1000 for a timeout value seems/is a bit excessive -- at least compared with the other timeout values in the test suite.

TheGeorge:
Since they seem to work now I set them to 1 min which should be plenty of time. From my understanding, the timeout does not behave well when it is specified for multiple tests as it was before.

test/exists_tests.erl
```diff
@@ -386,16 +399,51 @@ shrink3_test() ->
...
 prop_shrink4() ->
-  ?NOT_EXISTS(I, #{gen => integer(0, 20)},
+  ?NOT_EXISTS(I, integer(0, 20),
               begin
                 ?MINIMIZE(I),
                 I < 10
               end).
+
+matching_type() ->
+        ?LET(I, integer(), I).
```
kostis: Indent me properly, please.

ferd:
nice, had not seen this PR go through before! Thanks for the improvements!

PR #164, opened 2018-05-16
Adding Coverage Reports to PRs

TheGeorge:


TheGeorge:
@kostis  This doesn't do anything yet. I think we also need to collect coverage information and enable the codecov third party application. 

Take a look at https://github.com/parapluu/Concuerror/pull/253 for how coverage should be reported.

kostis:
Yes, I know.

PR #167, opened 2018-05-17
enable coverage reports for PRs

TheGeorge:
rebar produces coverage reports and covertool translates them into a format that covdecov should be able to understand.

Let's see if this works...


fenollp:
What coverage do you get with `rebar3 do cover —reset, eunit —cover, cover —verbose`? Cause I get coverage on only one module...

Note: — should be double dashes. I’m on a phone...

TheGeorge:
@fenollp PropEr uses rebar2, but you need to append 

```erlang
{cover_enabled, true}.
{cover_export_enabled, true}.
{cover_print_enabled, true}.
```

to `rebar.config` and then run `make tests` to get coverage reports.

TheGeorge:
So with this PR accepted we have one more travis build that produces coverage reports and shows them here. We should also be able to track coverage changes over time.

fenollp:
Nice! 
Note: rebar3 can totally be used to build & test PropEr.

kostis:
Note that there is no intention to include in the main repository anything that produces the tons of warnings that vanilla versions of rebar/rebar3 produce by default...

Long ago, we worked hard to create a version of rebar that suits PropEr's needs and that version will be substituted with another one _only_ if the newer one is _strictly_ superior.

TheGeorge:
So I now have a patched version of `rebar` that behaves like the one PropEr has in `master` right now. I guess I have to compile that version with OTP 17 for it to work on all Travis builds....

fenollp:
Not sure what warnings you’re talking about? Vanilla rebar3 does not produce many warnings for me. 

What is the rationale behind having your own rebar version in git too I’m curious? Usually it’s for apps meant to be standalone and PropEr is not this. Also erlang.mk supports rebar.config files fine nowadays. 
You can take a look at my other PR (branch rebaring) you’ll see it doesn’t produce useless warnings.

TheGeorge:
I rebased on master

TheGeorge:
Those files are already not included in the codecov report. It is only rebar who prints them out. If you want I can disable the text output of rebar.

kostis:
> Those files are already not included in the codecov report.

I am not sure where I can see the whole report.  I've only found something that shows a 71.14% coverage score down from 71.25% (which actually makes me a bit suspicious about these numbers since the only? diff between these two was the `rebar` change).

> If you want I can disable the text output of rebar.

I find the current version of the text output completely unnecessary, if not misleading (because I am not sure how/whether it corresponds to the codecov score), so I would vote to disable it.  No strong opinion though.  But it would be nice to have some clearly visible pointer to the actual codecov report with the _real_ percentages.

TheGeorge:
You click the `Continue to review full report at Codecov.` and then files and then you can inspect each file. 

I removed the rebar output and excluded the files that don't need coverage.

kostis:
> I removed the rebar output and excluded the files that don't need coverage.

I am not sure where the latter is shown.  For example, I see a lot of files with 0% coverage at (what I think is) the [latest report](https://codecov.io/gh/proper-testing/proper/tree/29443c74dfd97267947d13ce3724929d47d5fbe8/src).

If this is not the latest, can you please point me to it?

TheGeorge:
Yeah I'm not sure how that works with codecov. I think it still shows the
files because they where there in the first uploaded thingy and are still
files in the repo. I have to give this a closer look.

When I run the coverage tests locally the blacklisted files are not
included in the coverage testing so I think it has something to do with
codecov.

rebar.b6d3094.patch
```diff
@@ -0,0 +1,92 @@
...
+index 412b29d..1eec6bd 100644
+--- a/src/rebar_erlc_compiler.erl
++++ b/src/rebar_erlc_compiler.erl
+@@ -528,7 +528,9 @@ expand_file_names(Files, Dirs) ->
+         list()) -> ok | {ok, any()} | {error, any(), any()}.
+ internal_erl_compile(Config, Source, OutDir, ErlOpts) ->
+     ok = filelib:ensure_dir(OutDir),
+-    Opts = [{outdir, OutDir}] ++ ErlOpts ++ [{i, "include"}, return],
++    Opts = [{outdir, OutDir}] ++
++    [Opt || Opt <- ErlOpts, Opt =/=eunit_nowarnings]  ++
```
kostis: Add a space after `=/=`.

Also, in the next line, the code should probably read
`[{i, "include"} | [return || not lists:member(eunit_nowarnings, ErlOpts)]]`
instead (i.e., the `,` should be `|`).

rebar.b6d3094.patch
```diff
@@ -0,0 +1,92 @@
...
+                                                      ?EUNIT_DIR),
+     case CompileOnly of
+         "true" ->
+@@ -171,7 +172,12 @@ run_eunit(Config, CodePath, SrcErls) ->
+     StatusBefore = status_before_eunit(),
+     EunitResult = perform_eunit(Config, Tests),
+ 
+-    rebar_cover_utils:perform_cover(Config, FilteredModules, SrcModules,
++    BlackListedModules = rebar_config:get(Config, cover_skip_modules, []),
++    {CoverModules, CoverSrc} =
```
kostis: Wrapping these two in a tuple, instead of having two assignments in two lines, is totally unnecessary (and ugly).

PR #173, opened 2018-06-17
r3: simplify build

fenollp:
This is the build system simplification extracted from #168 
It makes use of the rebar `platform_define` to not require writing an HRL to the filesystem. This option is supported by rebar, rebar3 and the custom patched rebar that is bundled in this repo.

The idea is to simplify/standardize the build process so that one day a custom patched rebar would not need to be part of this repo as well as making development easier long term for contributors that use rebar3.


mix.exs
```diff
@@ -22,7 +22,7 @@ defmodule Proper.Mixfile do
   end
 
   defp package do
-     [files: ~w(src include rebar.config configure Makefile COPYING README.md THANKS check_escripts.sh clean_doc.sh clean_temp.sh write_compile_flags mix.exs),
+     [files: ~w(src include rebar.config configure Makefile COPYING README.md THANKS check_escripts.sh clean_doc.sh clean_temp.sh flags mix.exs),
```
kostis: Not sure what the `flags` file is.  Can you elaborate?

rebar.config
```diff
@@ -35,9 +37,22 @@
 {erl_opts, [debug_info,
 	    report_warnings, {warn_format,1}, warn_export_vars,
 	    warn_obsolete_guard, warn_unused_import,
-	    warn_missing_spec, warn_untyped_record]}.
+	    warn_missing_spec, warn_untyped_record,
+	    warn_unused_vars,
```
kostis: I thought that `warn_unused_vars` was on by default...  Do you have evidence to the contrary?

rebar.config
```diff
@@ -35,9 +37,22 @@
...
+	    {platform_define, "^2[1-9]", 'AT_LEAST_21'}]}.
 
-{pre_hooks, [{"(linux|darwin|gnu)", compile, "make include/compile_flags.hrl"},
-             {"(freebsd|netbsd|openbsd|solaris|dragonfly)", compile, "gmake include/compile_flags.hrl"},
-	     {"win32", compile, "escript.exe write_compile_flags include/compile_flags.hrl"}]}.
 {post_hooks, [{clean, "./clean_doc.sh"}]}.
+
+{dialyzer, [{warnings, [unmatched_returns
+                       ]}
+           ,{plt_extra_apps, [erts, kernel, stdlib, compiler, crypto, syntax_tools]}
```
kostis: Please follow the style of the rest of the file.

rebar.config
```diff
@@ -35,9 +37,22 @@
...
+           ]}.
+
+{profiles, [{test, [{erl_opts, [nowarn_missing_spec
```
kostis: Why are these `nowarn_` options present here? They should not really be, since this PR is still using (our patched version of) `rebar`, isn't it?

fenollp: This is rebar3-specific and clears great many a warnings when running test.

kostis: No, I do _not_ want to see that.  Instead, I want to see a pull request that does _one_ thing, and one thing only, and does it right.  This one is titled "Simplify build".  Let's have that.

kostis:
Please, do _not_ include in this PR changes which are rebar3-specific, if you actually want it merged.  I've made this point before but apparently it did not get across.  I hope it does now.

We want PRs that do one thing well and do one thing only.  If you want to also substitute `rebar` with `rebar3` open a new pull request.

fenollp:
Last commit has been removed. This looks ready to me ;)

kostis:
> Last commit has been removed. This looks ready to me ;)

Not to me though.  I still do not understand why the entry 
<pre>
{profiles, [{test, [{erl_opts, [nowarn_missing_spec
 ...
</pre>
appears in `rebar.config`.

According to your comment, this is rebar3-specific.  So it should also be removed.

fenollp:
Right. Just patched & squashed.

PR #175, opened 2018-07-12
Fix links in README.md and overview.edoc

aronisstav:


aronisstav:
Wait on this, i see some 'manopapad's too.

kostis:
> Should the following also be updated?

Yes, although this is relevant only for the next time we publish a new version on 'Hex.pm'.

aronisstav:
@kostis Done.

PR #177, opened 2018-07-28
Fix type signature of `USERNF` neighbour function.

ferd:
The doc specified the type `-type nf() :: fun((term(), temperature()) -> term()).`
but in practice, what the user is given contains the max temperature term as called in
https://github.com/proper-testing/proper/blob/b000cc92da9ccbe1764ba4494fea71fb614872f3/src/proper_gen_next.erl#L215-L217

### DIFF
--- a/src/proper_gen_next.erl
+++ b/src/proper_gen_next.erl
@@ -62,7 +62,7 @@
 -define(SLTEMP(T), adjust_temperature(T)).
 
 -type temperature() :: float().
--type nf() :: fun((term(), temperature()) -> term()).
+-type nf() :: fun((pos_integer(), temperature()) -> term()).
 -type matcher() :: fun((term(), proper_types:raw_type(), temperature()) -> term()).
 
 -spec update_caches('accept' | 'reject') -> 'ok'.
--- a/src/proper_target.erl
+++ b/src/proper_target.erl
@@ -53,7 +53,8 @@
 %%%   <dt>`?USERNF(Gen, Nf)'</dt>
 %%%   <dd>This uses the neighborhood function `Nf' instead of PropEr's
 %%%     constructed neighborhood function for this generator. The neighborhood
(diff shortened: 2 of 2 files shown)


TheGeorge:
Hi,
Sorry for coming back to this PR so late. The generated NFs can get internally called with a tuple `{Depth, Temperature}`. This functionality is there to support depth-dependent temperature scaling (in other words, the temperature can be calculated depending on how deep the current generator is nested in the overall generator structure).

I think the right way to fix this is to either change the type for `nf()`:
```
-type temperature() :: {pos_integer(), float()}.
```
or to wrap the user-defined NFs temperature argument so that the depth is not exposed:
```
set_user_nf(Type, NF) ->
  proper_types:add_prop(user_nf, fun (Base, T) -> NF(Base, ?TEMP(T)) end, Type).
```

ferd:
Yeah I guess the question is what, as a user of this library writing a custom function, should I be writing my custom neighbor selection against? So far the types have been wrong, but the actual arguments I've had were invariably the `{Depth :: number(), Temperature::float()}` argument, but I'm fine changing `Depth` to be `pos_integer()`. 

I'd be fine keeping the depth personally, since it can help toy a bit with a bunch of arguments, and it's always easy to ignore. Would you like me to amend the PR to use `pos_integer()` for depth instead of `number()`?

alfert:
Thanks for that clarification. I just started porting the targeted approach to Elixir as part of PropCheck and was asking me that particular question: what are the specs of the user provided neighbourhood function.

TheGeorge:
> I'd be fine keeping the depth personally, since it can help toy a bit with a bunch of arguments, and it's always easy to ignore. Would you like me to amend the PR to use pos_integer() for depth instead of number()?

Yes, you can fix the type for `nf()` and the change the docs accordingly.

ferd:
I've updated the type and the doc.

Issue #183, opened 2018-11-21
The `cant_generate` error could include info about the failing constraints


robertoaloi:
When the random instance generation subsystem fails to produce an instance that satisfies some `?SUCHTHAT` constraint, a `cant_generate` error is returned to the user. The error indicates that the user should either increase the `constraint_tries` limits or loosen the failing constraints, but it does not contain any reference to the actual constraints which are failing. A hint could be provided to the user by analyzing the failing constraints, instead. In fact, the failing `Type` is already available to the `proper_gen:generate/3` function (see https://github.com/proper-testing/proper/blob/master/src/proper_gen.erl#L165) which throws the `cant_generate` exception, but it's discarded. The `Type` could instead be inspected, the constraints extracted (e.g. via `proper_types:get_prop(constraints, Type)` and further analyzed to return useful information (e.g. the module and the function name) about the constraint funs causing the failure (e.g. via `erlang:fun_info(F)`.

kostis:
Thanks tor opening this issue!

Perhaps some concrete example(s) of failing `?SUCHTHAT` constraint generators and the info that one would like to see in that(those) case(s) would help us in doing something about that.

Even better, a PR on the subject is welcome!

robertoaloi:
@kostis I just opened a PR which shows an example of a failing `?SUCHTHAT` and compares the current output with a possible alternative. As a note, the reason why I included the whole _Type_ at the end of the output is because it sometimes contains some useful hints to identify the root cause of the issue (usually as part of the `env` property. Any feedback or comment is welcome.

kostis:
Fixed by #184

PR #184, opened 2018-11-24
Introduce a more informative cant_generate error

robertoaloi:
This PR attempts to address the issue #183 by introducing a more informative _cant_generate_ error in case of failure of the random instance generation subsystem due to a strict constraint (e.g. via a `?SUCHTHAT` macro). This is particularly useful for test modules defining a big number of constraints, where identifying the root cause of a failing constraint can be tricky. As a trivial example, given the following module, which contains an _impossible_ generator (`gen_2/0`):

```erlang
-module(test).

gen_1() ->
  ?SUCHTHAT(A, pos_integer(), A > 0).

gen_2() ->
  ?SUCHTHAT(B, pos_integer(), B < 0).

prop() ->
  ?FORALL({A, B}, {gen_1(), gen_2()}, A =:= B).
```

The current output from proper is:

```erl
2> proper:quickcheck(test:prop()).

Error: Couldn't produce an instance that satisfies all strict constraints after 50 tries.
{error,cant_generate}
```

This leaves the user in doubt if the error refers to the first or the second occorrence of `?SUCHTHAT`.

With the proposed patch, the root cause of the issue becomes more obvious:

```erl
Error: Couldn't produce an instance that satisfies all strict constraints from test:'-gen_2/0-fun-0-' after 50 tries.
{error,
    {cant_generate,
        {'$type',
            [{env,{1,inf}},
             {generator,{typed,#Fun<proper_types.13.103056237>}},
             {is_instance,{typed,#Fun<proper_types.14.103056237>}},
             {shrinkers,[#Fun<proper_types.15.103056237>]},
             {kind,basic},
             {constraints,[{#Fun<test.1.106658561>,true}]}]}}}
```

### DIFF
--- a/test/proper_tests.erl
+++ b/test/proper_tests.erl
@@ -983,7 +983,7 @@ false_props_test_() ->
      ?_fails(error_statem:prop_simple())].
 
 error_props_test_() ->
-    [?_errorsOut({cant_generate,{proper_tests,error_props_test_,0}},
+    [?_errorsOut({cant_generate,{?MODULE,error_props_test_,0}},
 		 ?FORALL(_, ?SUCHTHAT(X, pos_integer(), X =< 0), true)),
      ?_errorsOut(cant_satisfy,
 		 ?FORALL(X, pos_integer(), ?IMPLIES(X =< 0, true))),
@@ -1033,7 +1033,7 @@ options_test_() ->
      ?_failsWith([42], ?FORALL(_,?SHRINK(42,[0,1]),false), [noshrink]),
      ?_failsWith([42], ?FORALL(_,?SHRINK(42,[0,1]),false), [{max_shrinks,0}]),
      ?_fails(?FORALL(_,integer(),false), [fails]),
-     ?_assertRun({error,{cant_generate,{proper_tests,options_test_,0}}},
+     ?_assertRun({error,{cant_generate,{?MODULE,options_test_,0}}},
 		 ?FORALL(_,?SUCHTHAT(X,pos_integer(),X > 0),true),
 		 [{constraint_tries,0}], true),
      ?_failsWith([12],
--- a/test/proper_tests.erl
+++ b/test/proper_tests.erl
@@ -510,7 +510,8 @@ impossible_types() ->
      ?SUCHTHAT(B, binary(), lists:member(256,binary_to_list(B))),
      ?SUCHTHAT(X, exactly('Lelouch'), X =:= 'vi Brittania'),
      ?SUCHTHAT(X, utf8(), unicode:characters_to_list(X) =:= [16#D800]),
-     ?SUCHTHAT(X, utf8(1, 1), size(X) > 1)].
+     ?SUCHTHAT(X, utf8(1, 1), size(X) > 1),
+     ?SUCHTHAT(X, ?SUCHTHAT(Y, pos_integer(), Y > 0), X < 0)].
 
 impossible_native_types() ->
     [{types_test1, ["1.1","no_such_module:type1()","no_such_type()"]},
@@ -982,7 +983,7 @@ false_props_test_() ->
      ?_fails(error_statem:prop_simple())].
 
 error_props_test_() ->
-    [?_errorsOut({cant_generate,_},
+    [?_errorsOut({cant_generate,{proper_tests,error_props_test_,0}},
 		 ?FORALL(_, ?SUCHTHAT(X, pos_integer(), X =< 0), true)),
      ?_errorsOut(cant_satisfy,
 		 ?FORALL(X, pos_integer(), ?IMPLIES(X =< 0, true))),
@@ -992,8 +993,8 @@ error_props_test_() ->
 		   ?FORALL(X, integer(), ?IMPLIES(X > 5, X < 6))),
      ?_assertCheck({error,too_many_instances}, [1,ab],
 		   ?FORALL(X, pos_integer(), X < 0)),
-     ?_errorsOut({cant_generate,_}, prec_false:prop_simple()),
-     ?_errorsOut({cant_generate,_}, nogen_statem:prop_simple()),
+     ?_errorsOut({cant_generate,{proper_statem,commands,4}}, prec_false:prop_simple()),
+     ?_errorsOut({cant_generate,{nogen_statem,impossible_arg,0}}, nogen_statem:prop_simple()),
      ?_errorsOut(non_boolean_result, ?FORALL(_, integer(), not_a_boolean)),
      ?_errorsOut(non_boolean_result,
 		 ?FORALL(_, ?SHRINK(42,[0]),
@@ -1032,7 +1033,7 @@ options_test_() ->
      ?_failsWith([42], ?FORALL(_,?SHRINK(42,[0,1]),false), [noshrink]),
      ?_failsWith([42], ?FORALL(_,?SHRINK(42,[0,1]),false), [{max_shrinks,0}]),
      ?_fails(?FORALL(_,integer(),false), [fails]),
-     ?_assertRun({error,{cant_generate,_}},
+     ?_assertRun({error,{cant_generate,{proper_tests,options_test_,0}}},
 		 ?FORALL(_,?SUCHTHAT(X,pos_integer(),X > 0),true),
 		 [{constraint_tries,0}], true),
      ?_failsWith([12],
--- a/test/proper_tests.erl
+++ b/test/proper_tests.erl
@@ -511,7 +511,21 @@ impossible_types() ->
      ?SUCHTHAT(X, exactly('Lelouch'), X =:= 'vi Brittania'),
      ?SUCHTHAT(X, utf8(), unicode:characters_to_list(X) =:= [16#D800]),
      ?SUCHTHAT(X, utf8(1, 1), size(X) > 1),
-     ?SUCHTHAT(X, ?SUCHTHAT(Y, pos_integer(), Y > 0), X < 0)].
+     %% Nested constraints, of which the inner one fails
+     ?SUCHTHAT(X, ?SUCHTHAT(Y, pos_integer(), Y < 0), X > 0),
+     %% Nested constraints, of which the outer one fails
+     ?SUCHTHAT(X, ?SUCHTHAT(Y, pos_integer(), Y > 0), X < 0),
+     %% Nested constraints, one strict and one non-strict, where the inner one fails
+     ?SUCHTHATMAYBE(_X, ?SUCHTHAT(Y, pos_integer(), Y < 0), true),
+     %% Nested constraints, one strict and one non-strict, where the outer one fails
+     ?SUCHTHAT(X, ?SUCHTHATMAYBE(Y, pos_integer(), Y < 0), X < 0),
+     %% Two failing constraints within a ?LET macro, where one
+     %% constraint is used as the 'raw type' and one is used as the 'generator'
+     ?LET(Y,?SUCHTHAT(X, pos_integer(), X < 0),?SUCHTHAT(Y, pos_integer(), Y < 0)),
+     %% Two failing constraints within a ?LET macro, where both
+     %% constraints are used as a 'raw type'
+     ?LET({X,Y},{?SUCHTHAT(X1, pos_integer(), X1 < 0),?SUCHTHAT(Y1, pos_integer(), Y1 < 0)},{X,Y})
+    ].
 
 impossible_native_types() ->
     [{types_test1, ["1.1","no_such_module:type1()","no_such_type()"]},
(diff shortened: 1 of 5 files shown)


TheGeorge:
I agree, that the current output is not very verbose, but I think something
like line number and file or module would be even more helpful here.

robertoaloi:
Absolutely. This PR is just to start the discussion on the topic. I totally agree that line number + file would be clearer.

robertoaloi:
@TheGeorge Do you have a suggestion on how to get the line number for the failing constraint? Two alternatives that come to my mind are:

A. If the module is compiled with `debug_info`, one could use the `beam_lib` and `code` API to retrieve the line number. This sounds extremely ugly and fragile, though, and I would not go for it.
B. Modify the `SUCHTHAT*` macros and/or the `proper_types:add_constraint` function so that, when a constraint is added, we store a reference to the place where the constraint is defined (e.g. using the `?MODULE` and `?LINE` macros). This would be cleaner, but it potentially requires a bigger change.|

I'm open to suggestions.

TheGeorge:
I was thinking about options B. You could store this information as type property `info` or similar.

kostis:
First of all, thanks for opening this PR @robertoaloi !

Now, rather than focusing on "implementation details", such as how to get line info, I would like to first have a more general discussion on what is the _PropEr_ thing for this PR to do.

The original issue (#183) complained, rightfully so, that

> When the random instance generation subsystem fails to produce an instance that satisfies some `?SUCHTHAT` constraint, a `cant_generate` error is returned to the user. ... A hint could be provided to the user by analyzing the failing constraints, instead. ... The Type could instead be inspected, the constraints extracted (e.g. via proper_types:get_prop(constraints, Type) and further analyzed to return useful information about the constraints causing the failure.

So, ideally, the information that is returned to the user should _explain_ the reason(s) for failing to generate terms that satisfy the constraints, i.e., which constraints are unsatisfiable and/or conflicting. 
This is not done by the current PR.  Instead, the current PR:

1. Prints an error message which is better than the previous one.
2. Generates (and returns) a quite incomprehensible error term (the second element of the `cant_generate` tuple).

I think that point 1 above is good.  Perhaps it can be slightly improved by _massaging_ the name of the (function containing the) generator so that instead of printing:
```erl
Error: Couldn't produce an instance that satisfies all strict constraints from test:'-gen_2/0-fun-0-' after 50 tries.
```
it prints
```erl
Error: Couldn't produce an instance that satisfies all strict constraints from test:gen_2/0' after 50 tries.
```
Whether to include in the error message the line number of the function or not is a secondary issue, IMO.

As to point 2, I see very little reason to return PropEr's internal representation to the user -- in fact, arguably this is the wrong thing to do because it makes changing this representation more difficult to do in the future. So, this part of the PR needs to change. Ideally, it should provide the list of constraints that are unsatisfiable and/or difficult to satisfy. However, this may require "deep surgery in PropEr's code base" and/or be difficult to do accurately in all cases (running the risk of being more confusing to the user rather than helpful).  So, perhaps the term which is returned can be just:
```erlang
{cant_generate, {M,F,A}}
```
i.e., the MFA of the generator or the MFA where the generator is contained.


One more thing: the PR should contain more tests, perhaps the test in #183 and some others.  Also, if we go with my suggestion for returning the MFA as part of the `cant_generate` tuple, the existing tests of PropEr's test suite that were changed in this PR should be modified to match on the appropriate MFA (rather than matching on an `_`, as they currently do).

TheGeorge:
I agree, that printing the internal representation is not the right thing to do unless we can somehow pretty-print the constraint. 

The problem is that PropEr and it's macros don't work with MFAs but with funs. So returning `{cant_generate, {M, F, A}}` will not work. I think the line number and the file are would be a good starting point for locating which constraint fails. Right now it is very difficult to locate the constraint since next to no information is given.

kostis:
> The problem is that PropEr and it's macros don't work with MFAs but with funs.

Yes, I know that, but funs in Erlang are _always_ associated with the MFA where they are defined.

For example, take a look at the error message that is printed by this PR. The `test:'-gen_2/0-fun-0-'` fun contains all information about the MFA where it is defined.  It's both possible and easy to reconstruct the MFA from the fun's name.

TheGeorge:
Ah nice, then it works with MFA. I still would like to have the line
number. It gives some additional information (e.g. if multiple constraints
are defined in one function) and is just easy to use.

robertoaloi:
Hi again.

I totally agree on dropping the internal representation of the type from the value 'returned' to the user. It was there only to show you what was available and what could potentially be extracted, but I see no point in showing that to the user in the final solution. This is also the reason why I used a wildcard for the term in the tests for the time being.

I like Kostis' suggestion to return a MFA, since it's a good compromise between having no information at all and having the file and line number (which I agree is the ideal end state of this, but may require a lot of more work for little advantage). The reason why I didn't attempt at parsing the fun name is that I didn't know how stable the fun naming schema was in Erlang but, after a little research, it looks like it has been quite stable. There's actually code in OTP which already uses the same strategy to fetch the 'parent', so it looks like it's safe to do what Kostis suggested:

https://github.com/erlang/otp/blob/423f078e3b8603cbd8bfea3be8709092a861a36e/lib/eunit/src/eunit_lib.erl#L379

I will also try to extend the tests a bit. One of Kostis' comments above contains an important keyword:

> So, ideally, the information that is returned to the user should explain the reason(s) for failing to generate terms that satisfy the constraints, i.e., which constraints are unsatisfiable and/or **conflicting**.

I want to see how the new code will behave (i.e. add a test) in the case of 'conflicting' constraints. I also need to check other constraints such as the ones produced by `?SUCHTHATMAYBE`, which may (or may not) reserve some surprise.

kostis:
@robertoaloi : Sounds like a very good plan.  Looking forward to it!

robertoaloi:
Then I'll proceed with the changes.

It looks like the `?SUCHTHATMAYBE` is not an issue since it never results in a constraint failure (by definition). Also, in the case of multiple failing constraints, it looks like we can probably focus only on the first one and ignore the others. Or would you prefer to be safe and return/deal with all the parents from the constraint MFAs?
Finally, whilst we could just the API from `eunit_lib` to fetch the parent, I'm not sure it's worth to create a dependency from proper to eunit just for this. How do you feel about that?

kostis:
> Also, in the case of multiple failing constraints, it looks like we can probably focus only on the first one and ignore the others.

Focus on the first one.  This also makes the return value always having just one MFA instead of an MFA list.

> Finally, whilst we could just the API from eunit_lib to fetch the parent, I'm not sure it's worth to create a dependency from proper to eunit just for this.

I would prefer no dependency. Just copy the code and leave a comment that this comes from `eunit_lib`.

test/proper_tests.erl
```diff
@@ -982,7 +983,7 @@ false_props_test_() ->
      ?_fails(error_statem:prop_simple())].
 
 error_props_test_() ->
-    [?_errorsOut({cant_generate,_},
+    [?_errorsOut({cant_generate,{proper_tests,error_props_test_,0}},
```
kostis: `?MODULE` should be better here instead of `proper_tests`.
Same for line 1036 below.

robertoaloi:
It looks like I need to amend the code from `eunit_lib` to ensure backward compatibility.

kostis:
The `fun_parent` function exists in previous OTP versions.  For example, in 17.0 its code looks as follows
```erlang
fun_parent(F) ->
    {module, M} = erlang:fun_info(F, module),
    {name, N} = erlang:fun_info(F, name),
    case erlang:fun_info(F, type) of
        {type, external} ->
            {arity, A} = erlang:fun_info(F, arity),
            {M, N, A};
        {type, local} ->
            [$-|S] = atom_to_list(N),
            C1 = string:chr(S, $/),
            C2 = string:chr(S, $-),
            {M, list_to_atom(string:sub_string(S, 1, C1 - 1)),
             list_to_integer(string:sub_string(S, C1 + 1, C2 - 1))}
    end.
```
Perhaps you can factor out the parts that need to differ before and after 20.x and use a code similar to:
```erlang
-ifdef(AT_LEAST_20).
...   % use string:split here
-else.
...
-endif.
```
the `AT_LEAST_20` macro is/should be available to all modules by the `rebar.config` file.

kostis:
On second thought, perhaps calling `eunit_lib:fun_parent` directly will do the trick.

robertoaloi:
I saw your comments only now, after I included my latest changes. I just pushed a version of the `fun_parent/1` function which should work for all recent versions of OTP, even if it's not the most beautiful one.

If I had to pick between the `AT_LEAST_20` macro and the library call I would vote for the library call. Let me know if you are happy with the current function or if you'd like me to instead invoke the library.

TheGeorge:
> > Also, in the case of multiple failing constraints, it looks like we can probably focus only on the first one and ignore the others.

> 
> Focus on the first one. This also makes the return value always having just one MFA instead of an MFA list.

Is this correct though? When generators have multiple constraints, we cannot be sure that the first one is the one that fails. We might report the MFA for the wrong constraint.

kostis:
I am not sure I get the point of your example.  Have you tried it?

My understanding is that we will (and should) return `{cant_generate, {Mod, myint, 0}}` in this case.

kostis:
@robertoaloi wrote
> If I had to pick between the AT_LEAST_20 macro and the library call I would vote for the library call. Let me know if you are happy with the current function or if you'd like me to instead invoke the library.

I gave it a bit of a thought and I think that the chances that `eunit_lib:parent_fun/1` will disappear or do something totally different at some future point are quite slim.  Let's use that instead, rather than duplicating it in a way that is different both from that of Erlang/OTP 17.0 and of 20.0.

When you do that change, please make sure that you also add `eunit` in the list of applications in dialyzer's PLT (this is in the `Makefile` of PropEr).

Once some more tests have been added, I will merge this PR.

robertoaloi:
@TheGeorge The approach works because the `Condition` specified in the `?SUCHTHAT` is wrapped in a `fun` by the macro and that fun is the 'constraint' which gets added to the 'type'.

@kostis Agreed on using the library.

robertoaloi:
Actually, @TheGeorge 's intuition is correct, even if the counter-example provided above is not. See the following one, instead:

```
-module(test).
-compile(export_all).

possible() ->
  ?SUCHTHAT(X, pos_integer(), X > 0).

impossible() ->
  ?SUCHTHAT(X, possible(), X < 0).

impossible_gen() ->
  ?LET(Impossible, impossible(), Impossible).

impossible_prop() ->
  ?FORALL(_X, impossible_gen(), true).
```

This would return:

```
{error,{cant_generate,{test,possible,0}}}
```

Which is incorrect.

The point is that for some reason I assumed that the boolean contained in the constraint was an indication of a failing constraint, but my assumption was incorrect. That boolean flag simply indicates if the constraint is strict or not. So, I don't think we can return a single MFA.

robertoaloi:
I modified the code, the respective tests and the doc so that a list of candidate MFAs is returned on `cant_generate`. I also added a few extra tests. If you have any other scenario that you would like to test, please advise.

TheGeorge:
Looks good to me

kostis:
@robertoaloi : Thanks again for the PR!

I've reviewed the code and overall it looks good.  The only line that looks a bit weird is the following one in the tests:
```erlang
  ?_errorsOut({cant_generate,[{proper_statem,commands,4}]}, prec_false:prop_simple()),
```
What's happening in this test is that a list of commands cannot be generated because some precondition (defined in the `prec_false` module) is false. Still, the `cant_generate` error which is returned mentions only the `proper_statem:commands/4` generator, which is actually nowhere mentioned in the `prec_false` module.  This may be confusing (to a user).

Can you check whether something can be done for this case too?  Otherwise, I will merge it as is.

robertoaloi:
@kostis I've been thinking about this as well.

Whilst it's true that the `proper_statem:commands/4`generator is not directly referred from the `prec_false` module, the `proper_statem:commands/1` **is** (even if not via its fully qualified name because of the `import`) and that should already provide a precious hint to the user. Compared to the previous implementation, this basically tells the user that the problem is not in one of her generators, but it's related to the generation of command sequences for proper. Understanding that this has to do with _impossible_ preconditions is a bit of a hard leap, but seems a step towards the right direction. Maybe we could simply add a line in the doc (either next to the `cant_generate`doc or in the `statem` section) to explain things a bit better.

In other words, I don't see any obvious solution to improve this other than modifying the `proper_statem` code  or making a special case (read, a horrible hack) for the `proper_statem` module. I'd avise against both approaches. Instead, I vote for merging this PR as is (maybe with a doc addition) and, potentially, improve things further in a separate PR, where we could add some metadata such as line number (and maybe a _message_ in the case of the statem) when adding a constraint to the type.

kostis:
> Instead, I vote for merging this PR as is (maybe with a doc addition) and, potentially, improve things further in a separate PR, where we could add some metadata such as line number (and maybe a message in the case of the statem) when adding a constraint to the type.

Agreed.  Do you want to have the honor to add something appropriate to the docs, or should I press the merge button and handle it myself separately?

robertoaloi:
Let me give it a try.

robertoaloi:
@kostis I found a sweet spot in the `proper_statem` where to add a reference to the error. I didn't feel like adding it to the `proper` module documentation because that's agnostic of the statem. The PR is ready from my perspective. Let me know if you want me to squash the 7 commits into one. Otherwise we are ready to go!

kostis:
Looks good.

No need to squash the commits -- I will hit the GitHub magic button that does this once Travis finishes.

PR #189, opened 2019-02-25
Publish edoc to hexdocs

aaronrenner:
It would be ideal if proper's edoc based documentation was published to hexdocs, instead of the current empty ExDocs based documentation:

screen shot 2019-02-24 at 9 42 35 pm

I know it's possible to upload edoc to hexdocs because it's currently done for sbroker here: https://hexdocs.pm/sbroker/.

The [documentation for `mix hex.publish`](https://hexdocs.pm/hex/Mix.Tasks.Hex.Publish.html) provides the following info.

> Documentation will be generated by running the `mix docs` task. `ex_doc` provides this task by default, but any library can be used. Or an alias can be used to extend the documentation generation. The expected result of the task is the generated documentation located in the `doc/` directory with an `index.html` file.

This PR adds a `docs` alias to your mix file so when you run `mix hex.publish` or `mix hex.publish docs`, it should automatically to hexdocs. I can't test fully because I don't have publishing permissions, but I can confirm when I run `mix hex.publish docs` it does automatically generate the edoc for the project.

I also removed the `ex_doc` dependency since doesn't seem like it's needed anymore with this change. Thanks for all of your great work on this library!

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -6,13 +6,12 @@ defmodule Proper.Mixfile do
      version: "1.3.0",
      description: description(),
      package: package(),
-     deps: deps()]
+     deps: deps(),
+     aliases: aliases()]
   end
 
   defp deps do
-    [
-      {:ex_doc, "~> 0.14", only: :dev},
-    ]
+    []
   end
 
   defp description do
@@ -27,4 +26,8 @@ defmodule Proper.Mixfile do
      licenses: ["GPL"],
      links: %{"GitHub" => "https://github.com/proper-testing/proper"}]
    end
+
+  defp aliases do
+    [docs: "cmd ./make_doc"]
+  end
 end


kostis:
Thanks for your interest in PropEr and for your PR.

I cannot test this either, so I need to make a decision whether to merge this based on the hope that it works and it proves interesting for the Elixir community.

However, out of curiosity, wouldn't it be easier to just add a link(links) from hexdocs to the ["official" PropEr API documentation](https://proper-testing.github.io/apidocs/), esp. since it is now on GitHub, rather than duplicating the information there and run the risk to have versions that are out of sync or not up-to-date?  

```
<disclaimer>
I've never seen hexdocs before to know its advantages.  The above is just a thought.
</disclaimer> 
```

aaronrenner:
One of the key advantages for me is that anything published in hexdocs can be downloaded locally and is fully searchable within the [Dash documentation browser](https://kapeli.com/dash). This makes it seamless to jump between looking up the APIs for [PropCheck](https://github.com/alfert/propcheck) and Proper.

I figured out how to run a test to ensure the docs publish as expected.

1. Renamed app to `hexdocs_proper` locally (you can call it whatever you like).
2. `mix hex.publish`
3. Opened up hexdocs and confirmed the edoc-based docs were uploaded properly.
    screen shot 2019-02-25 at 7 20 55 am
4. Unpublished the package with `mix hex.publish --revert 1.3.0`. **This must be done within an hour or an admin needs to get involved.**

Please let me know what you think.

kostis:
OK, if you claim it works, I will merge this.  Thanks again for the PR.

aaronrenner:
Thank you so much! ❤️

If you do run into an issue, feel free to mention me with @aaronrenner and I'd be happy to help.

PR #201, opened 2019-08-22
Convert targeted to proper interface

xspirus:
This pull request was done with 2 problems in mind:

1. Working to implement the Stateful TPBT of PropEr, I came across a problem with `?FORALL_TARGETED`, which turned out to come from not accepting `?TRAPEXIT` as a valid test. The problem was caused by [this line of code](https://github.com/proper-testing/proper/blob/3d74fde55450609bf11ba37bb28835375a2a5708/include/proper_common.hrl#L38), which does not accept any "non-immediate" boolean properties.
2. When calling the `?FORALL_TARGETED` macro, the messages after passing or failing are not in line with the rest of the PropEr interface, that is what was printed was always that the tests passed or failed "After 1 test(s).".

The changes made to the code, do not have any big implications on the `?FORALL_TARGETED` functionality, rather aim to fix the problems stated above. That is, `?FORALL_TARGETED` now is executing the number of tests defined by `numtests` and accepts "non-immediate" boolean properties.

### DIFF
--- a/include/proper_common.hrl
+++ b/include/proper_common.hrl
@@ -35,7 +35,7 @@
 -define(EXISTS(X,RawType,Prop), proper:exists(RawType, fun(X) -> Prop end, false)).
 -define(NOT_EXISTS(X,RawType,Prop), proper:exists(RawType, fun(X) -> Prop end, true)).
 -define(FORALL_TARGETED(X, RawType, Prop),
-        proper:exists(RawType, fun(X) -> not Prop end, true)).
+        proper:targeted(RawType, fun(X) -> Prop end)).
 -define(IMPLIES(Pre,Prop), proper:implies(Pre,?DELAY(Prop))).
 -define(WHENFAIL(Action,Prop), proper:whenfail(?DELAY(Action),?DELAY(Prop))).
 -define(TRAPEXIT(Prop), proper:trapexit(?DELAY(Prop))).
--- a/src/proper.erl
+++ b/src/proper.erl
@@ -387,7 +387,7 @@
 -export([get_size/1, global_state_init_size/1,
 	 global_state_init_size_seed/2, report_error/2]).
 -export([pure_check/1, pure_check/2]).
--export([forall/2, exists/3, implies/2, whenfail/2, trapexit/1, timeout/2, setup/2]).
+-export([forall/2, targeted/2, exists/3, implies/2, whenfail/2, trapexit/1, timeout/2, setup/2]).
 
 -export_type([test/0, outer_test/0, counterexample/0, exception/0,
 	      false_positive_mfas/0, setup_opts/0]).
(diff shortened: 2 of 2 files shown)


TheGeorge:
Hi @xspirus, can you give some insight into what you want to achieve with this change? As in what is the behavioral difference before and after the diff and why do we need to do it.
To me it looks mostly like code-duplication from the `?FORALL` path. It also does not seem to use the `perform_search`.

Also, could you please merge all commits into one if they don't contain individual working states?

xspirus:
Hello @TheGeorge . So the main reason of this modification is, as I am working on `TPBT` for stateful systems with @kostis , to make `?FORALL_TARGETED` compatible with stateful testing as well. The main problem was that `Prop` in the case of stateful testing is not always a `boolean` but sometimes wrapped in `?TRAPEXIT` and `?WHENFAIL`. So the first modification comes as a need to accomodate stateful `TBPT`.

The second part of the changes, are mostly to comply with the rest of the `proper` interface, as `?FORALL_TARGETED` did not take into account `numsteps` and it also did not display the number of tests executed (always printed as 1 test).

TheGeorge:
Thanks for coming back. Good to know someone picked up the targeted testing
l. :-)

Some context as to why things are implemented as they are. The targeted
constructs (including  ?EXISTS) use the config parameter search steps
instead of numsteps. This was done to make them behave better when
combining with the  untargeted variants. For example ?FORALL(..., ...,
?EXISTS(..., ..., ...)). Does not restart the search from the inner ?EXISTS
while ?FORALL(..., ..., ?FORALL(...)). conceptually does create new values
the whole way through every step.

For the second problem with non-boolean but valid return value I think
there is value in fixing it for all targeted variants, as they should be
working there as well.

xspirus:
Hi again @TheGeorge , thanks for offering insight on why things are implemented the way they are.

So the changes I made only impact the `?FORALL_TARGETED` macro, while the `?EXISTS` macro remains intact. I think that `?FORALL_TARGETED` is not meant to be used combined with other macros (as there are the `?EXISTS` macros). Essentially, the `?FORALL_TARGETED` macro is a "nicer" version of `?EXISTS` to be used on its own, thus the need to show the number of tests etc.

As for fixing for non-boolean values, I think that removing the negation and reverting the `Not` value to its opposite should be sufficient. What do you think?

kostis:
Let me add my .02 here.

I think there are various things that we can learn from how this PR was done.
1. PRs should come with descriptions on what they do and/or improve; a title only, let alone a slightly cryptic one, does not suffice.
2. If applicable, and I think that it is applicable in this case, PRs should also contain test(s) that demonstrate the new/better functionality they offer.
3. I think that the code of PropEr in general, and the targeted portion of it in particular, desperately needs some documentation that explains "why things are implemented as they are".  Now, of course I understand that writing documentation is boring and is not top priority for developers, but when situations like these arise, which involve aspects of the system that are not obvious (here I am referring to @TheGeorge's comment "This was done to make them behave better when combining with the untargeted variants, ..."), we should be adding this information _in the code_ and, ideally, also add some tests that ensure/demonstrate the functionality that "the current way of implementing things" intends to preserve/offer.

Regarding what to do about the points above, here is my suggestion:
1. A PR description would be nice to have here, but perhaps it's not necessary anymore because most information exists in the comments that this PR contains.
2. Tests definitely need to be added.  Minimally, a test that shows what this PR improves, but also a  test that demonstrates how "valid, but non-boolean return value(s)" can now be properly handled. (Although perhaps this may need the extension for stateful TBP to be in place first. If so, the appropriate test can be added as part of this PR and only be enabled when the extension is in place.)  Finally, some tests that `?FORALL(..., ..., ?EXISTS(..., ..., ...))` still work as they did/should also need to be added.
3. We should document some "non-obvious" design decisions.

@xspirus : Can you handle 2 (mainly) and 1 above?
@TheGeorge : Can you add 3?

PS. As to merging all commits, I think this is not necessary. One can inspect the changes in the "Files changed" button on GitHub, and we can squash the commits when merging.

kostis:
Small update:

The test suite already has tests that contain `?FORALL(..., ..., ?EXISTS(..., ..., ...))` combinations and these pass.  Tests that check that we can now handle valid non-boolean return values will be added as part of the `?TRAPEXIT` support of targeted PBT.  I've reviewed the rest of the changes and they look OK to me.  I can merge this PR once @TheGeorge reviews and approves it.

PR #211, opened 2019-10-15
Build PropEr using rebar3 instead of own rebar

kostis:
This set of changes allows PropEr to be built using rebar3 instead of
its own (patched) rebar version.  Besides being more modern, this
allows for easier integration of PropEr in code bases that use it as
a dependency.

Note that PropEr's version of rebar is still used for running the unit
tests, mainly due to being able to shut off more warnings than when
using rebar3.  However, this only affects PropEr developers, not users.

Addresses #204 and a request by propcheck.

Obsoletes #44 and most of #168.


alfert:
For packaging with hex, following the tutorial (https://www.rebar3.org/docs/publishing-packages), 
please add this to `app.src` (values taken from `mix.exs`): 

```erlang
     {maintainers, ["Manolis Papadakis", "Eirini Arvaniti", "Kostis Sagonas"]},
     {licenses, ["GPL3"]},
     {link,  [{"GitHub", "https://github.com/proper-testing/proper"}]},
```

According to https://www.rebar3.org/docs/using-available-plugins#section-hex-package-management the hex plugin for rebar is configured globally for in `~/.config/rebar3/rebar.config`: 
```erlang
{plugins, [rebar3_hex]}.
```

Publishing the package to hex can be done with `rebar3 hex publish` and `rebar3 hex docs`. 

@kostis, this was the information I grabbed together from rebar docs and the mix file. Shall we put it into a separate PR?

.gitignore
```diff
@@ -8,6 +8,8 @@ doc/*.css
 doc/*.html
 doc/*.png
 doc/edoc-info
-.directory
-.eunit
-.rebar
+rebar3
+rebar.lock
```
TheGeorge: This file is typically checked in.

Makefile
```diff
@@ -30,15 +30,19 @@ endif
...
 
 all: compile dialyzer doc test
 
 compile:
-	$(PROPER_REBAR) compile
+	$(REBAR3) compile
 
 dialyzer: .plt/proper_plt compile
-	dialyzer -n -nn --plt $< -Wunmatched_returns ebin
+	dialyzer -n -nn --plt $< -Wunmatched_returns _build/default/lib/proper/ebin
```
TheGeorge: rebar3 has a dialyzer command that should take care of building the plt and including the correct beams

.gitignore
kostis: Why should it be?  PropEr does not have any dependencies, let alone locked ones, and it seems that `rebar3` automatically generates this `rebar.lock` file when run.  What I am missing?

Makefile
kostis: I know about the `rebar3 dialyzer` command but I do not like it.  It requires having an installed Erlang/OTP system (which I do not have) in order to find `erts` as an application to include in the PLT.  I prefer calling `dialyzer` directly, and I do not see the point in not using `make` for these things.  Note that calling Dialyzer and running the tests is for PropEr developers only, not for regular users.

kostis:
Thanks @TheGeorge for the comments, but note that this PR is already merged. (It's of course possible to improve on it but it needs to happen in another PR.)

We want you to comment on #212 instead :-)

PR #212, opened 2019-10-15
Cleanup Targeted Implementation

xspirus:
This PR is intended to remove some code in the `proper_sa` implementation, where the `state` was represented as a `dict`. However, after checking all the tests, as well as the examples, it seemed that this `dict` is always of size 1, hence its existence is not necessary.

This change should also act as an optimization to the `targeted` implementation, as there is no more iteration of the `dict`.

### DIFF
--- a/include/proper_common.hrl
+++ b/include/proper_common.hrl
@@ -75,7 +75,7 @@
 %% Macros for backwards compatibility
 %%------------------------------------------------------------------------------
 
--define(TARGET(TMap), proper_target:targeted(make_ref(), TMap)).
+-define(TARGET(TMap), proper_target:targeted(TMap)).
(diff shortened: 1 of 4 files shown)


TheGeorge:
Thanks for cleaning this up! Once upon a time, there it was intented to have multiple targets in the generator, with a way to individual optimize them. (minimize and maximize another). However, this was never used in practice, so it can as well be dropped.

TheGeorge:
Preferably, you should squash both commits into one, they are both small and make only sense together.

kostis:
Thanks @xspirus for the PR and @TheGeorge for the review.

I've followed the reviewer's advice and merged this after squashing the commits.

PR #213, opened 2019-10-16
Add rebar3-hex plugin

evnu:
This pull request is a follow-up to #211 and https://github.com/alfert/propcheck/issues/121. The pull request adds the [rebar3_hex](https://github.com/tsloughter/rebar3_hex) plugin to allow publishing to hex.pm.  If the user is already registered with hex.pm, publishing boils down to two steps:

* `rebar3 hex user auth` (authenticate with hex.pm)
* `rebar3 hex publish` (publish to public hex.pm)

Note that compiling with `rebar` from the repository itself results in a warning, but the build completes:

```
./rebar compile
WARN:  Missing plugins: [rebar3_hex]
==> proper (compile)
...
```

@kostis Can you CC the people responsible for publishing to hex.pm? I tried the plugin using a small example project, but did not dare to try `rebar3 hex publish` within PropEr itself.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -45,3 +45,5 @@
 
 {dialyzer, [{warnings, [unmatched_returns]},
             {plt_extra_apps, [erts, kernel, stdlib, compiler, crypto, syntax_tools]}]}.
+
+{plugins, [rebar3_hex]}.


kostis:
Thanks @evnu -- I will merge this.

Regarding the warning on `./rebar compile`, I do not worry about it because it's not my intention to be using PropEr's own `rebar` for compiling from now on.  However, `./rebar eunit`, which is still used (until I find out a way to shut off the warnings of the test suite when using `rebar3 unit`), also produces the same warning.  Anyway, I hope it's something only temporary.

Regarding notifying the people responsible to hex.pm, I think they get these PR notifications, but I will cc: them anyway (hi @ferd!).  However, note that PropEr version 1.4.0 is not ready to be published yet.  It will be once you see me changing the badge / README on GitHub.

ferd:
You actually don't want this plugin, now everyone using PropEr gets to download the hex plugin that way. It should instead go in `{project_plugins, [rebar3_hex]}` if it is intended to only be used by project owners.

Generally we advise that only project owners use the hex plugin to publish, and that can be done using a global plugin rather than the local one. In any case, we publish [the rebar3_hex package](https://hex.pm/packages/proper) with mix so that it is better supported for Elixir folks without having to specify the `manager: :rebar3` config value.

I can publish new versions of PropEr to hex without a problem, but I would revert this commit first, since it will likely "pollute" builds and not add anything required for users of the library.

evnu:
@ferd Thanks for clarifying this!

kostis:
I've created a PR (#214) that reverts this -- also, see a comment there.

PR #214, opened 2019-10-16
Revert "Add rebar3-hex plugin"

kostis:
Reverts proper-testing/proper#213

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -45,5 +45,3 @@
 
 {dialyzer, [{warnings, [unmatched_returns]},
             {plt_extra_apps, [erts, kernel, stdlib, compiler, crypto, syntax_tools]}]}.
-
-{plugins, [rebar3_hex]}.


kostis:
Continuing the discussion of #213, I will revert this.

Should I also add a `{project_plugins, [rebar3_hex]}` to `rebar.config` or not?  (This part is not so clear to me.)

Kostis, who is (slowly) entering the brave `rebar3` world...

ferd:
No need to, that plugin is something I tend to maintain globally. 

Project plugins are more useful for example when people install the rebar3 plugin for proper I have written (https://github.com/ferd/rebar3_proper) which lets you expose the `rebar3 proper` command when using your app at the root, but won't include it when the app is included as a dep (and therefore won't make `rebar3 proper` available to the app that includes the dep that uses PropEr)

PR #216, opened 2019-10-17
Cleanup tests

kostis:
In order to eliminate unnecessary compiler warnigns.


kostis:
I was hoping I would add more in this PR, but half a year has passed and did not find time for it.

Perhaps this PR can be merged as is -- the grandiose plan to have completely a warning-free test suite and switch to `rebar3` for running the tests will have to be implemented later ;-)

PR #218, opened 2019-10-27
make: Manually restore ebin dir after rebar3 compile

kostis:


xspirus:
Should `ebin` now be placed in `.gitignore`?

PR #220, opened 2019-10-28
Use gen_server for proper_targeted

### LINKED ISSUE #219: ?FORALL_TARGETED does not work with ?TRAPEXIT
### Bug

While in a previous PR #201 , there was a fix so that `?FORALL_TARGETED` accepted properties with `?TRAPEXIT`. However, after experimenting with the targeted interface, I came across a bug referring these two macros when used in conjuction.

### Example

```erlang
-module(trapexit).

-export([prop_normal_targeted/0, prop_trapexit_targeted/0]).

nf() ->
    fun(Prev, _T) ->
        integer(Prev, inf)
    end.

gen() ->
    ?USERNF(integer(), nf()).

prop_normal_targeted() ->
    ?FORALL_TARGETED(I, gen(), 
                     begin
                         ?MAXIMIZE(I),
                         I < 1000
                     end).

prop_trapexit_targeted() ->
    ?FORALL_TARGETED(I, gen(),
                     ?TRAPEXIT(
                        begin
                            ?MAXIMIZE(I),
                            I < 1000
                        end)).
```

Opening this program in an erlang shell we have the following:

```
1> c(trapexit).
{ok, trapexit}
2> proper:quickcheck(trapexit:prop_normal_targeted()).
................................................!
Failed: After 49 test(s).
1005

Shrinking .(1 time(s))
1000
false
3> proper:quickcheck(trapexit:prop_trapexit_targeted()).
....................................................................................................
OK: Passed 100 test(s).
true
```

It seems that when a property is wrapped within a `?TRAPEXIT`, the targeted generation is not working properly.

### Problem Source

The problem seems to be related with the way that `proper_sa.erl` is implemented. Specifically, it uses the process dictionary to store the current state of the targeted search. So when used in conjuction with `?TRAPEXIT`, which spawns a new process, the progress of the search is not actually saved in the main process.

### Solution

I have come up with a solution, which is just replacing the process dictionary with the use of a public `ets` table, in the `proper_sa.erl` implementation. I will create a Pull Request later so that it can be reviewed.

xspirus:
Implement `proper_target` with a `gen_server`.
The callbacks for the `proper_target` behaviour have been updated as well.

Fixes #219.
There one can see a detailed explanation of a bug and the motivation for this change.

### DIFF
--- a/examples/level.erl
+++ b/examples/level.erl
@@ -168,7 +168,7 @@ path() ->
 
 path_next() ->
   fun (PrevPath, _) ->
-    ?LET(NextSteps, vector(20, step()), PrevPath ++ NextSteps)
+      ?LET(NextSteps, vector(20, step()), PrevPath ++ NextSteps)
   end.
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
--- a/examples/level.erl
+++ b/examples/level.erl
@@ -168,7 +168,7 @@ path() ->
 
 path_next() ->
   fun (PrevPath, _) ->
-      ?LET(NextSteps, vector(20, step()), PrevPath ++ NextSteps)
+    ?LET(NextSteps, vector(20, step()), PrevPath ++ NextSteps)
   end.
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
@@ -188,19 +188,19 @@ prop_exit_targeted(LevelData) ->
   #{entrance := Entrance} = Level,
   #{exit := Exit} = Level,
   ?FORALL_TARGETED(Path, ?USERNF(path(), path_next()),
-		   case follow_path(Entrance, Path, Level) of
-		       {exited, _Pos} -> false;
-		       Pos ->
-			   case length(Path) > 500 of
-			       true ->
-				   proper_sa:reset(),
-				   true;
-			       _ ->
-				   UV = distance(Pos, Exit),
-				   ?MINIMIZE(UV),
-				   true
-			   end
-		   end).
+                   case follow_path(Entrance, Path, Level) of
+                     {exited, _Pos} -> false;
+                     Pos ->
+                       case length(Path) > 500 of
+                         true ->
+                           proper_target:reset(),
+                           true;
+                         _ ->
+                           UV = distance(Pos, Exit),
+                           ?MINIMIZE(UV),
+                           true
+                     end
+                   end).
 
 distance({X1, Y1}, {X2, Y2}) ->
   math:sqrt(math:pow(X1 - X2, 2) + math:pow(Y1 - Y2, 2)).
--- a/include/proper_common.hrl
+++ b/include/proper_common.hrl
@@ -66,7 +66,7 @@
 %% Target macros
 %%------------------------------------------------------------------------------
 
--define(MAXIMIZE(Fitness), proper_target:update_target_uvs(Fitness, inf)).
+-define(MAXIMIZE(Fitness), proper_target:update_uv(Fitness, inf)).
 -define(MINIMIZE(Fitness), ?MAXIMIZE(-Fitness)).
 -define(USERNF(Type, NF), proper_gen_next:set_user_nf(Type, NF)).
 -define(USERMATCHER(Type, Matcher), proper_gen_next:set_matcher(Type, Matcher)).
@@ -77,7 +77,8 @@
 
 -define(TARGET(TMap), proper_target:targeted(TMap)).
 -define(STRATEGY(Strat, Prop), ?SETUP(fun (Opts) ->
-                                          proper_target:use_strategy(Strat, Opts),
+                                          proper_target:init_strategy(
+                                            Opts#{search_strategy => Strat}),
                                           fun proper_target:cleanup_strategy/0
                                       end, Prop)).
 -define(FORALL_SA(X, RawType, Prop),
--- a/include/proper_common.hrl
+++ b/include/proper_common.hrl
@@ -77,8 +77,8 @@
 
 -define(TARGET(TMap), proper_target:targeted(TMap)).
 -define(STRATEGY(Strat, Prop), ?SETUP(fun (Opts) ->
-                                          proper_target:init_strategy(
-                                            Opts#{search_strategy => Strat}),
+                                          NewOpts = Opts#{search_strategy => Strat},
+                                          proper_target:init_strategy(NewOpts),
                                           fun proper_target:cleanup_strategy/0
                                       end, Prop)).
 -define(FORALL_SA(X, RawType, Prop),
--- a/src/proper_sa.erl
+++ b/src/proper_sa.erl
@@ -28,81 +28,178 @@
 %%% that can be used to address discrete and continuous optimization problems.
 %%%
 %%% SA starts with a random initial input. It then produces a random input in
-%%% the neighborhood of the previous one and compares the fitnessof both. If
+%%% the neighborhood of the previous one and compares the fitness of both. If
 %%% the new input has a higher fitness than the previous one, it is accepted
 %%% as new best input. SA can also accepts worse inputs with a certain
-%%% probbability.
+%%% probability.
 %%% (<a target="_blank" href="https://en.wikipedia.org/wiki/Simulated_annealing">more information</a>)
 
 -module(proper_sa).
 
 -behaviour(proper_target).
 
-%% callbacks
--export([init_strategy/1,
-         init_target/1,
-         cleanup/0,
-         store_target/1,
-         retrieve_target/0,
-         update_fitness/1,
-         get_shrinker/1
-        ]).
-%% lib
--export([reset/0, get_last_fitness/0]).
-
 -include("proper_internal.hrl").
 
-%% macros and configuration parameters
--define(REHEAT_THRESHOLD, 5).
--define(RESTART_THRESHOLD, 100).
+%% -----------------------------------------------------------------------------
+%% Exports
+%% -----------------------------------------------------------------------------
+
+-export([init_strategy/1, init_target/1, next/2, update_fitness/3, reset/2]).
+
+%% -----------------------------------------------------------------------------
+%% Macros
+%% -----------------------------------------------------------------------------
 
 -define(RANDOM_PROBABILITY, (?RANDOM_MOD:uniform())).
 
--define(SA_DATA, proper_sa_data).
--define(SA_REHEAT_COUNTER, proper_sa_reheat_counter).
--define(SA_TABLE, proper_sa_table).
+%% -----------------------------------------------------------------------------
+%% Types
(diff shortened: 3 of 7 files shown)


TheGeorge:
@xspirus thanks again for cleaning up the targeted stuff in PropEr. Using ETS as a storage is surely more performant. However, the direct usage of ETS like this to have something persistent is discouraged. It's the Erlang equivalent of having a global variable. 

It's much better to encapsulate the access behavior in a gen_server that manages the ETS (and cleans it up when it dies/terminates). That server can be named and linked against the proper root process.

xspirus:
Hello @TheGeorge .

Thanks for the reply. Why is it discouraged to use `ETS` in this manner?

So it would be better to use a `gen_server` implementation where the state is the `sa_data` record?

If I understand correctly the flow would be this:

- Init strategy
  - Spawn `gen_server`
- Perform search
- Clean strategy
  - Stop `gen_server`

Am I correct?

fenollp:
Implementing a storage strategy as a module also adds the benefit of making it possible to support different storage kinds.
* a `gen_server` process that owns an ETS table
* a driver process to a distributed storage solution
* a `gen_server` process that uses its process dictionary (ETS really can't be faster than Proc.Dict, can it?)
* ... any module that implements the usual Erlang CRUD behavior

Could you add an option (could be compile-time) which one would use to specify what storage module to use?

TheGeorge:
The storage strategy probably does not have a huge influence on the overall performance. I don't think it's necessary to parameterise it. TBH, if I would redo the whole thing again, I'd probably implement the whole proper_sa module as a gen_server holding the data in the server state. The API just calls out to the server.

ETS is a bit overkill, as there is no parallel access. Sure, process dictionary is fast, but ugly to use. I believe a map is more than sufficient and has a way better ergonomic/readability.

TheGeorge:
> 
> Am I correct?

yes :-)

xspirus:
> yes :-)
Great

If we were to make a re-implementation of the `proper_sa`, then shouldn't the `proper_target` code be updated as well? Some stuff are still implemented with the thought of having multiple searches going on in the same time (in #212 I just removed the `make_ref()` usages).

TheGeorge:
@xspirus I just took a look at the `proper_target.erl` to fresh up my memory. 

Actually, following the standard gen_whatever design pattern, that thing should be the gen_server using callbacks from proper_sa. It should also handle storing the state and just forward it to the strategy implementation. 

This gets maybe a bit out of scope for fixing the bug, but would definitely be the cleanest (closest to OTP design patterns) solution.

xspirus:
@TheGeorge However, if `proper_target` is responsible for keeping track of the state, it should come with an "agnostic" manner. As the targeted works right now, as far as I can understand, `proper_target` should be an interface which can use different strategies (regardless if the only one available is `proper_sa` at the moment).

My point is that for example `simulated annealing` needs to keep track of temperature and other algorithm specific variables, while another strategy could need other variables. So if the state is handled by `proper_target` then it needs to be able to handle different types of states, which I do not know if is a good practice.

TheGeorge:
@xspirus look at how gen_server works. The callback module does not take care of storing and retrieving the state. This would work very similar: proper_target keeps the state and handles the access (ets, process dictionary, gen_server) and forwards it to the search strategy. proper_sa just defines how the state looks like, but not how it is stored. This has the advantage of replacing the underlying storage solution, if necessary, without touching any of the strategy code. 

So proper_sa would still define what it saved, but proper_target how it is saved (plus a common api towards the rest of proper)

xspirus:
@TheGeorge you are right, my bad. I forgot that `gen_server` can have anything as a state :)

kostis:
I have another comment here, which is slightly unrelated to the above discussion.
Besides whatever fix we agree on, this PR should also include the code of issue #219 as a test.

xspirus:
@kostis I will add that as well.

xspirus:
@TheGeorge thank you for your comments. I updated the codebase with the recommended changes.

src/proper_target.erl
```diff
@@ -61,70 +61,149 @@
...
+%% @doc Initializes targeted gen server based on a search strategy.
+%% Depending on the argument, the search steps will be taken from
+%% from the process dictionary or the setup options.
+
+-spec init_strategy(Opts :: opts()) -> {ok, pid()}.
+init_strategy(Strat) when is_atom(Strat) ->
+  Strategy = strategy(Strat),
+  put('$search_strategy', Strategy),
+  Steps = get('$search_steps'),
+  proper_gen_next:init(),
```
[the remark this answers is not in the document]
xspirus: Yes, this should be done as a separate PR maybe.

src/proper_target.erl
```diff
@@ -151,57 +223,74 @@ strategy(Strat) ->
...
+  {NextValue, NewTarget, NewData} = Strat:next(Target, Data),
+  {reply, NextValue, State#state{target = NewTarget, data = NewData}}.
 
 %% @private
--spec update_target(target_state()) -> 'ok'.
-update_target(State) ->
-  Strategy = strategy(),
-  {_, N, F} = Strategy:retrieve_target(),
-  Strategy:store_target({State, N, F}).
+-spec handle_cast(Request :: term(), State :: state()) ->
```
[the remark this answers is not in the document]
xspirus: All calls are now synchronous

src/proper_target.erl
```diff
@@ -61,70 +61,159 @@
...
+
+-spec init_strategy(Opts :: opts()) -> {ok, pid()}.
+init_strategy(Strat) when is_atom(Strat) ->
+  Steps = get('$search_steps'),
+  init_strategy(#{search_steps => Steps, search_strategy => Strat});
+init_strategy(#{search_steps := Steps, search_strategy := Strat}) ->
+  Strategy = strategy(Strat),
+  proper_gen_next:init(),
+  Data = Strategy:init_strategy(#{numtests => Steps}),
+  gen_server:start_link({local, ?MODULE}, ?MODULE, [{Strategy, Data}], []).
```
fenollp: Is this really intended to be a singleton? Wouldn't two concurrent calls to `quickcheck` corrupt this gen_server's state?

xspirus: This could be fixed by using the servers `pid`, if I am not wrong. In this case however the `?MAXIMIZE` macro would be broker, as it has no way of knowing the `pid` of the current `proper_target gen_server`. 

I don't know if there is a solution to this.

TheGeorge: Yes, PropEr stores all run specific information there anyways and that makes it consistent and parallel runs work as well.

TheGeorge:
LGTM
Thanks for the cleanup!

src/proper_target.erl
TheGeorge: Just saw that you already did this :-)

xspirus: @TheGeorge Great. I checked that the same pattern was followed with the typeserver, so I already made the necessary changes.

xspirus:
@TheGeorge thanks for the review and the suggestions. Glad I'm able to help!

src/proper_target.erl
```diff
@@ -61,70 +61,171 @@
...
+reset() ->
+  TargetserverPid = get('$targetserver_pid'),
+  safe_call(TargetserverPid, reset).
+
+%% Create a safe call to a gen_server in case it
+%% raises noproc. Τhis should only be used for
+%% calls that do not return significant values.
+
+%% @private
+-spec safe_call(pid(), term()) -> term() | ok.
```
kostis: The `| ok' here is superfluous.  I suggest you take it out.

examples/level.erl
```diff
@@ -168,7 +168,7 @@ path() ->
 
 path_next() ->
   fun (PrevPath, _) ->
-      ?LET(NextSteps, vector(20, step()), PrevPath ++ NextSteps)
```
kostis: I think that this is the default indentation that Emacs gives to `fun`s, so perhaps it should not be changed.

include/proper_common.hrl
```diff
@@ -77,7 +77,8 @@
 
 -define(TARGET(TMap), proper_target:targeted(TMap)).
 -define(STRATEGY(Strat, Prop), ?SETUP(fun (Opts) ->
-                                          proper_target:use_strategy(Strat, Opts),
+                                          proper_target:init_strategy(
+                                            Opts#{search_strategy => Strat}),
```
kostis: I would prefer this line being written in two lines as follows:
```erlang
NewOpts = Opts#{search_strategy => Strat},
proper_target:init_strategy(NewOpts),
```

src/proper_sa.erl
```diff
@@ -28,80 +28,178 @@
...
+
+%% The function which generates the next instances of
+%% the targeted generator. It also updates the target state.
+
+%% @private
+-spec next(sa_target(), sa_data()) -> {any(), sa_target(), sa_data()}.
+next(Target, Data) ->
+  Temperature = Data#sa_data.temperature,
+  NextGenerator = (Target#sa_target.next)(Target#sa_target.last_generated,
+                                          Temperature),
```
kostis: The four lines above are a bit ugly.  I suggest the following rewrite:
```erlang
next(#sa_target{next = Next, last_generated = LastGen} = Target, Data) ->
  NextGenerator = Next(LastGen, Data#sa_data.temperature),
```
which is both prettier and half the size (in lines).

src/proper_sa.erl
```diff
@@ -28,80 +28,178 @@
...
+-spec update_fitness(proper_target:fitness(), sa_target(), sa_data()) ->
+  {sa_target(), sa_data()}.
+update_fitness(Fitness, Target, Data) ->
+  #sa_data{k_current = K_Current,
+           k_max = K_Max,
+           last_energy = Energy,
+           temperature = Temperature,
+           temp_func = TempFunc} = Data,
+  case (Energy =:= null)
+       orelse (Data#sa_data.p)(Energy,
```
kostis: Why isn't the `Data#sa_data.p` field taken out as part of the record deconstruction above?
if you change that, then most likely the whole case will fit in one line, instead of being across 4 of them!

src/proper_target.erl
```diff
@@ -61,70 +61,171 @@
...
+%% @doc Cleans up proper_gen_next as well as stopping the gen_server.
+
+-spec cleanup_strategy() -> ok.
+cleanup_strategy() ->
+  case erase('$targetserver_pid') of
+    undefined -> ok;
+    TargetserverPid ->
+      proper_gen_next:cleanup(),
+      ok = gen_server:stop(TargetserverPid),
+      ok
```
kostis: This `ok` line can be removed.  Similarly for the line two lines below it.

src/proper_target.erl
```diff
@@ -151,57 +245,76 @@ strategy(Strat) ->
...
+handle_call(gen, _From, State) ->
+  Strat = State#state.strategy,
+  Target = State#state.target,
+  Data = State#state.data,
+  {NextValue, NewTarget, NewData} = Strat:next(Target, Data),
+  {reply, NextValue, State#state{target = NewTarget, data = NewData}};
+handle_call({init_target, TMap}, _From, State) ->
+  Strat = State#state.strategy,
+  Target = State#state.target,
+  NewTarget = case Target of
```
kostis: The indentation of this `case` is not what you would get by Emacs...   Is this a problem?

xspirus: I think the new identation is correct now.

Issue #224, opened 2020-03-25
run_parallel_commands doesn't handle exceptions in commands the same way run_commands does


x4lldux:
When a command crashes during `run_parallel_commands`, the entire test just crashes also, while`run_commands` catches such a situation and returns a `exception` tuple as a result. Shouldn't they work the same way?

kostis:
Please submit a minimal test case that illustrates the issue.

x4lldux:
I can, but it'll be in Elixir using PropCheck library (a thin wrapper around PropEr); my Erlang is not that strong. But I've traced the origin of it:
- `run_parallel_commands` uses [`pmap`](https://github.com/proper-testing/proper/blob/master/src/proper_statem.erl#L624) to run commands in parallel, wrapped in `execute`,
- `execute` is [using `apply`](https://github.com/proper-testing/proper/blob/master/src/proper_statem.erl#L667) BIF instead of `safe_apply` like `run_commands` does
- `pmap` spawns process and awaits the answers, 
- spawned process catch any error, while executing the command, and send it as a [message](https://github.com/proper-testing/proper/blob/master/src/proper_statem.erl#L656) to `pmap`
- when the message, `pmap` gets, is a crash it just [rethrows it](https://github.com/proper-testing/proper/blob/master/src/proper_statem.erl#L667)

So when a SUT has a crashes, `run_commands` will return `{exception, _, _, _}` tuple as a result, while `run_parallel_commands` just forwards the crash.

kostis:
> I can, but it'll be in Elixir using PropCheck library (a thin wrapper around PropEr); my Erlang is not that strong.

Well, you've came to the PropEr (TM) repository.  Here you have your chance to strengthen your Erlang skills ;-)

x4lldux:
> Well, you've came to the PropEr (TM) repository.

Which runs on a BEAM™ that supports many other languages. Here you have your chance to make us feel included ;)

x4lldux:
@kostis test case added, see commit x4lldux/proper@c932eef62cfddc8a2726d09a2cb9bf5e81fd6015

x4lldux:
Also, the docs for `run_parallel_commands` are missing the exception tuple as a valid result type and that's what is being returned when command raises during the sequential part.

PR #227, opened 2020-03-26
Parallel crash fix, #224

x4lldux:
WIP, my attempt of fix for #224


x4lldux:
@kostis it's no longer WIP and is ready for review.

kostis:
Thanks for the #224 issue and this PR. If all goes well, I will look into this before the end of the week.

In the meantime, you may want to correct an erroneous `any` that should read `term()` in the return of `execute/3`,

kostis:
OK, it took longer than expected because in the meantime I decided to first [increase the test suite coverage](https://github.com/proper-testing/proper/commit/272b211292dbf05fdc55cc167f032a39fecb9136) and then [to fix two statem tests to actually test what they are supposed to](https://github.com/proper-testing/proper/commit/c3d485cff4a2aaba7382b84a83b2391711c608bd), but now I have looked at this.

Long story short: I do not (fully) understand the issue and I have doubts that the test shows what you think it shows.  I would like to see a test where a property like this one
```erlang
prop_exception() ->
    ?FORALL(Cmds, commands(?MODULE),
	    begin
		{_History,_State,Result} = run_commands(?MODULE, Cmds),
		Result =:= ok
	    end).
```
fails because `Result` is not `ok` but an `{exception,_,_,_}` tuple, while this property
```erlang
prop_parallel_exception() ->
    ?FORALL(Cmds, parallel_commands(?MODULE),
	    begin
		{_History,_State,Result} = run_parallel_commands(?MODULE, Cmds),
		Result =:= ok
	    end).
```
does not do what it is supposed to because of an exception, but it does with the PR that fixes the issue reported in #241.

Note the crucial difference in what I am suggesting and what this PR contains as test: `run_parallel_commands/2` is supposed to be run with a pair containing a list of lists of commands in its 2nd element (*as produced by `parallel_commands/1`*) while the 2nd arg of `run_commands/2` takes as input just a list of commands (as produced by `commands/1`).

Please supply the test case first that shows what goes wrong without this PR.

PR #228, opened 2020-03-26
Fix more_commands/2

### LINKED ISSUE #192: more_commands/2 does not generate more commands
The size of the generated commands for a state machine via `commands/1` seems to depend on the `maximum_size`, as expected. According to the docs using `more_commands/2` should help here and generate more commands. But this does not seem to work. 

Example code: 
```erlang
-module(prop_sizedbug).
-compile(export_all).

initial_state() -> ok.
    
command(_S) ->
    oneof([{call,?MODULE,foo,[integer()]},
	   {call,?MODULE,bar,[]}]).

precondition(_, _) -> true.
next_state(_, _, _) -> ok.
postcondition(_, _, _) -> true.

foo(_I) -> ok.
bar() -> 42.

prop_simple() ->
    ?FORALL(Cmds, commands(?MODULE),
        measure("Length of Cmds", length(Cmds), length(Cmds) >= 0)).

prop_more() ->
    ?FORALL(Cmds, more_commands(20, commands(?MODULE) ),
        measure("Length of Cmds", length(Cmds), length(Cmds) >= 0)).
```

Result from running `rebar3 proper` is 
```
===> Testing prop_sizedbug:prop_simple()
....................................................................................................
OK: Passed 100 test(s).

Length of Cmds
minimum: 0
average: 10.34
maximum: 36
===> Testing prop_sizedbug:prop_more()
....................................................................................................
OK: Passed 100 test(s).

Length of Cmds
minimum: 0
average: 10.18
maximum: 38
===> 
2/2 properties passed
```

For `prop_more()` I expected that something upto 800 commands would be generated in the command list. Setting the `max_size` does helps here, but still `more_commands/2` has no effect.

x4lldux:
Original `more_commands/2` wasn't working as expected because the adjusting of
size with `resize/2` wasn't propagating to "deeper" types.

Fixes #192.

With these changes I get:

```
....................................................................................................
OK: Passed 100 test(s).

Length of Cmds
minimum: 0
average: 8.93
maximum: 36
....................................................................................................
OK: Passed 100 test(s).

Length of Cmds
minimum: 0
average: 96.33
maximum: 398
```

### DIFF
--- a/src/proper_statem.erl
+++ b/src/proper_statem.erl
@@ -238,6 +238,8 @@
 -define(WORKERS, 2).
 -define(LIMIT, 12).
 
+-define(COMMANDS_SZ_FACTOR, '$commands_size_factor').
+-define(RESIZE_FACTOR, proper_types:parameter(?COMMANDS_SZ_FACTOR, 1)).
 
 %% -----------------------------------------------------------------------------
 %% Exported only for testing purposes
@@ -308,9 +310,10 @@ commands(Mod) ->
 	 ?SUCHTHAT(
 	    Cmds,
 	    ?LET(List,
-		 ?SIZED(Size,
-			proper_types:noshrink(
-			  commands(Size, Mod, InitialState, 1))),
+             ?SIZED(Size,
+                    proper_types:noshrink(
+                      commands(Size * ?RESIZE_FACTOR,
+                               Mod, InitialState, 1))),
 		 proper_types:shrink_list(List)),
 	    is_valid(Mod, InitialState, Cmds, []))).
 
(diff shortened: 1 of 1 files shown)


src/proper_statem.erl
```diff
@@ -308,9 +308,12 @@ commands(Mod) ->
 	 ?SUCHTHAT(
 	    Cmds,
 	    ?LET(List,
-		 ?SIZED(Size,
-			proper_types:noshrink(
-			  commands(Size, Mod, InitialState, 1))),
+                 ?LET(CommandsSizeFactor,
+                      proper_types:parameter('$commands_size_factor', 1),
```
kostis: I am wondering about the purpose of the `?LET` here...
Long lines aside, wouldn't it be simpler to skip the `?LET` and write the above as
```erlang
             ?SIZED(Size,
￼                   proper_types:noshrink(
￼                       commands(Size * proper_types:parameter('$commands_size_factor', 1),
￼                                Mod, InitialState, 1)))),
```
instead?

(Similar change in the code below, of course.)

x4lldux: You're right. I've pushed the changes.

kostis:
Thanks @x4lldux for your PR!

As you can see, based on your change, I've done a (IMO) stylistically better change to `proper_statem` (in the commit which is referenced above) and also added two related tests in [this branch](https://github.com/kostis/proper/tree/more_commands_fix).

You can either change your commit to have the changes of [this one](https://github.com/kostis/proper/commit/d5a238f458c7b042f3b8cdb217f48ab5c7e80eff) and then I can merge this PR -- so that you get the PropEr credit -- or alternatively I can close this PR and push the changes of my branch.

Let me know which you prefer.  In either case, your name (or github account ID if you do not wish to reveal your name) can be added to [this file](https://github.com/proper-testing/proper/blob/master/THANKS).

PR #233, opened 2020-03-30
Fix targeted shrinking in USERNF

### LINKED ISSUE #221: No Shrink in TPBT
While experimenting with TPBT properties, I came across some cases where shrinking was not happening at all for the generators, when using the `?USERNF` macro.

## Example

```erlang
-module(bad_shrink).
-compile(export_all).

%% -----------------------------------------------------------------------------
%% Generators
%% -----------------------------------------------------------------------------

normal_int() ->
  integer().

let_int() ->
  ?LET(I, integer(), I).

nf_int() ->
  fun (Prev, _T) ->
    ?LET(I, integer(Prev, inf), I)
  end.

normal_int_user_nf() ->
  ?USERNF(normal_int(), nf_int()).

let_int_user_nf() ->
  ?USERNF(let_int(), nf_int()).

normal_list() ->
  list(integer()).

let_list() ->
  ?LET(L, list(integer()), L).

nf_list() ->
  fun (Prev, _T) ->
    Max = case Prev of
      [] -> 0;
      _ -> lists:max(Prev)
    end,
    ?LET(L, vector(length(Prev), integer(Max, inf)), Prev ++ L)
  end.

normal_list_user_nf() ->
  ?USERNF(normal_list(), nf_list()).

let_list_user_nf() ->
  ?USERNF(let_list(), nf_list()).

%% -----------------------------------------------------------------------------
%% Properties
%% -----------------------------------------------------------------------------

property_int(I) ->
  ?MAXIMIZE(I),
  I < 500.

prop_normal_int() ->
  ?FORALL_TARGETED(I, normal_int_user_nf(), property_int(I)).

prop_let_int() ->
  ?FORALL_TARGETED(I, let_int_user_nf(), property_int(I)).

property_list(L) ->
  ?MAXIMIZE(lists:sum(L)),
  lists:sum(L) < 1000.

prop_normal_list() ->
  ?FORALL_TARGETED(L, normal_list_user_nf(), property_list(L)).

prop_let_list() ->
  ?FORALL_TARGETED(L, let_list_user_nf(), property_list(L)).
```

### Output

```
1> c(bad_shrink).
bad_shrink.erl:3: Warning: export_all flag enabled - all functions will be exported
{ok,bad_shrink}
2> proper:quickcheck(bad_shrink:prop_normal_int()).
.....................!
Failed: After 22 test(s).
513

Shrinking .(1 time(s))
500
false
3> proper:quickcheck(bad_shrink:prop_let_int()).
..............................................!
Failed: After 47 test(s).
559

Shrinking (0 time(s))
559
false
4> proper:quickcheck(bad_shrink:prop_normal_list()).
..!
Failed: After 3 test(s).
[-2,-7,0,-1,3,-9,21,29,8,3,5,6,29,33,29,32,35,33,35,33,37,33,30,29,46,38,37,46,38,50,71,45,49,39,45,38,49,39,40,42,47,44,41,43,42,44,52,39]

Shrinking ...(3 time(s))
[20,37,46,38,50,71,45,49,39,45,38,49,39,40,42,47,44,41,43,42,44,52,39]
false
5> proper:quickcheck(bad_shrink:prop_let_list()).
...!
Failed: After 4 test(s).
[23,8,34,36,44,45,49,48,50,49,55,50,55,52,56,54,61,60,67,60,56,60,57,60,71,75,56,64,62,61,59,59]

Shrinking (0 time(s))
[23,8,34,36,44,45,49,48,50,49,55,50,55,52,56,54,61,60,67,60,56,60,57,60,71,75,56,64,62,61,59,59]
false
```

## Problem

The problem seems to come from the way `?USERNF`s are treated when shrinking. Specifically in [`proper.erl`](https://github.com/proper-testing/proper/blob/a0222a73527ecadcee9472fecc8b1cbaa56547a8/src/proper.erl#L1776) where if the generator is a `user_nf` the instance is cleaned. This works well if the `first` generator is of a simple type (`integer()`, `float()`, etc.), but not so well if both generators `first` and `next` produce same instances (not simple).

## Solution

Lines 1776-1781 in `proper.erl` could become:

```erlang
ImmInstance = case proper_gen:safe_is_instance(RawImmInstance, Type) of
  false -> proper_gen:clean_instance(RawImmInstance);
  _ -> RawImmInstance
end,
```

However, this would not solve opposite cases, where the `first` generator is complex (`?LET` for example) and the `next` function is a simple generator.

xspirus:
This commit fixes #221 and updates the way the targeted implementation
shrinks its instances. PropEr now tries to find the best shrinker
between the original generator and the neighborhood function.

Users can now also specify `?SHRINK` macros in the neighborhood
function, which will take precedence.

### DIFF
--- a/src/proper.erl
+++ b/src/proper.erl
@@ -387,7 +387,8 @@
 -export([get_size/1, global_state_init_size/1,
 	 global_state_init_size_seed/2, report_error/2]).
 -export([pure_check/1, pure_check/2]).
--export([forall/2, targeted/2, exists/3, implies/2, whenfail/2, trapexit/1, timeout/2, setup/2]).
+-export([forall/2, targeted/2, exists/3, implies/2,
+         whenfail/2, trapexit/1, timeout/2, setup/2]).
 
 -export_type([test/0, outer_test/0, counterexample/0, exception/0,
 	      false_positive_mfas/0, setup_opts/0]).
@@ -445,15 +446,12 @@
 		      | {'setup', setup_fun(), outer_test()}
 		      | {'numtests', pos_integer(), outer_test()}
 		      | {'on_output', output_fun(), outer_test()}.
--type targeted_type() :: 'none'
-                       | proper_types:type().
 %% TODO: Should the tags be of the form '$...'?
 %% @type test(). A testable property that has not been wrapped with an
 %% <a href="#external-wrappers">external wrapper</a>.
 -opaque test() :: boolean()
 	        | {'forall', proper_types:raw_type(), dependent_test()}
 	        | {'exists', proper_target:tmap(), dependent_test(), boolean()}
-                | {'targeted', proper_target:tmap(), targeted_type(), dependent_test()}
 	        | {'conjunction', [{tag(),test()}]}
 	        | {'implies', boolean(), delayed_test()}
 	        | {'sample', sample(), stats_printer(), test()}
@@ -955,8 +953,6 @@ peel_test({setup,Fun,OuterTest}, #opts{setup_funs = Funs} = Opts) ->
     peel_test(OuterTest, Opts#opts{setup_funs = [Fun|Funs]});
 peel_test({exists,_,_,_} = ExistsTest, Opts) ->
     {ExistsTest, Opts#opts{numtests=1}};
-peel_test({targeted, _, _, _} = TargetedTest, #opts{numtests = NumTests} = Opts) ->
-    {TargetedTest, Opts#opts{search_steps = NumTests}};
 peel_test(Test, Opts) ->
     {Test, Opts}.
 
@@ -1002,9 +998,15 @@ exists(RawType, DTest, Not) ->
     {exists, #{gen => RawType}, DTest, Not}.
 
 %% @private
--spec targeted(proper_types:raw_type(), dependent_test()) -> test().
+-spec targeted(proper_types:raw_type(), dependent_test()) -> outer_test().
 targeted(RawType, DTest) ->
-    {targeted, #{gen => RawType}, none, DTest}.
+  setup(fun(#{numtests := Numtests} = Opts) ->
+            NewOpts = Opts#{search_steps => Numtests},
+            proper_target:init_strategy(NewOpts),
+            proper_target:init_target(#{gen => RawType}),
+            fun proper_target:cleanup_strategy/0
+        end,
+        forall(proper_target:targeted(#{gen => RawType}), DTest)).
 
 %% @doc Returns a property that is true only if all of the sub-properties
 %% `SubProps' are true. Each sub-property should be tagged with a distinct atom.
@@ -1224,14 +1226,6 @@ get_rerun_result(#fail{}) ->
 get_rerun_result({error,_Reason} = ErrorResult) ->
     ErrorResult.
 
--spec perform(pos_integer(), test(), opts()) -> imm_result().
-perform(NumTests, {targeted, TMap, _Target, Prop}, #opts{search_strategy = Strat} = Opts) ->
-    proper_target:init_strategy(Strat),
-    Target = proper_target:targeted(TMap),
-    Res = perform(0, NumTests, ?MAX_TRIES_FACTOR * NumTests,
-                  {targeted, TMap, Target, Prop}, none, none, Opts),
-    proper_target:cleanup_strategy(),
-    Res;
 perform(NumTests, Test, Opts) ->
     perform(0, NumTests, ?MAX_TRIES_FACTOR * NumTests, Test, none, none, Opts).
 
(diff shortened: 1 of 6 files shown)


include/proper_common.hrl
```diff
@@ -34,8 +34,7 @@
 -define(FORALL(X,RawType,Prop), proper:forall(RawType, fun(X) -> Prop end)).
 -define(EXISTS(X,RawType,Prop), proper:exists(RawType, fun(X) -> Prop end, false)).
 -define(NOT_EXISTS(X,RawType,Prop), proper:exists(RawType, fun(X) -> Prop end, true)).
--define(FORALL_TARGETED(X, RawType, Prop),
-        proper:targeted(RawType, fun(X) -> Prop end)).
```
kostis: I think this was written this way because we do not want lines > 80 chars

xspirus: Yes will update with fixes as well.

src/proper_target.erl
```diff
@@ -265,6 +274,12 @@ handle_call(gen, _From, State) ->
   Data = State#state.data,
   {NextValue, NewTarget, NewData} = Strat:next(Target, Data),
   {reply, NextValue, State#state{target = NewTarget, data = NewData}};
+handle_call(shrinker, _From, State) ->
+  Strat = State#state.strategy,
```
kostis: These three lines that access these three fields can be written in just one.

src/proper.erl
```diff
@@ -1774,12 +1751,14 @@ shrink(Shrunk, [ImmInstance | Rest], {_Type,Prop}, Reason,
...
-		      {ok, _} ->
-			  proper_gen:clean_instance(RawImmInstance);
-		      error ->
-			  RawImmInstance
-		  end,
+    ImmInstance = case get('$user_nf') of
+                    true ->
+                      case proper_types:safe_is_instance(RawImmInstance, Type) of
+                        false -> proper_gen:clean_instance(RawImmInstance);
+                        _ -> RawImmInstance
```
kostis: this `_' should be `true`, I think, unless you really want it to match with `{error,{typeserver,...}}` that `safe_is_instance` also returns.

src/proper.erl
```diff
@@ -1774,12 +1751,14 @@ shrink(Shrunk, [ImmInstance | Rest], {_Type,Prop}, Reason,
...
+                      end;
+                    _ -> RawImmInstance
```
kostis: Similarly to the above, the `_` here should read `false` for avoiding this case matching if `$user_nf` is not set.

xspirus: This could be `undefined` if `$user_nf` is not set at all.

src/proper.erl
```diff
@@ -1774,12 +1741,15 @@ shrink(Shrunk, [ImmInstance | Rest], {_Type,Prop}, Reason,
...
-		      error ->
-			  RawImmInstance
-		  end,
+    ImmInstance = case proper_types:find_prop(is_user_nf, Type) of
+                    {ok, true} ->
+                      case proper_types:safe_is_instance(RawImmInstance, Type) of
+                        false -> proper_gen:clean_instance(RawImmInstance);
+                        true -> RawImmInstance
+                      end;
+                    %% Either {ok, false} or error, suggests no user nf.
```
kostis, line 1750: I think it is better that this code is written in one of the following two ways:
```erlang
    Other when Other =:= {ok, false} orelse Other =:= erorr -> RawImmInstance
```
or
```erlang
   {ok, false} -> RawImmInstance;
   error -> RawImmInstance
```
because then it is robust w.r.t. what `proper_types:find_prop` returns.

src/proper_sa.erl
```diff
@@ -76,33 +77,36 @@
...
--record(sa_target,
-        {first             = null :: proper_types:type(),
-         next              = null :: fun((_, _) -> proper_types:type()),
-         current_generated = null :: proper_gen:instance(),
-         last_generated    = null :: proper_gen:instance()}).
+-record(
+   sa_target,
+   {first             = null :: proper_types:type(),
+    next              = null :: fun((_, _) -> proper_types:type()
+                                                | proper_gen:instance()),
```
kostis, line 84: I think it's better to declare a type for `proper_types:type() | proper_gen:instance()` and use it here.

This will also allow to undo the change in the formatting of the two record, which adds, unnecessarily, many lines of differences in this PR.

src/proper_target.erl
```diff
@@ -258,16 +281,17 @@ init([{Strategy, Data}]) ->
 %% @private
 -spec handle_call(Request :: term(), From :: {pid(), Tag :: term()},
                   State :: state()) ->
-  {reply, Reply :: term(), NewState :: state()}.
-handle_call(gen, _From, State) ->
-  Strat = State#state.strategy,
-  Target = State#state.target,
-  Data = State#state.data,
+        {reply, Reply :: term(), NewState :: state()}.
+handle_call(gen, _From,
```
kostis, line 285: I think this looks better as:
```erlang
handle_call(gen, _From, State) ->
  #state{strategy = Strat, target = Target, data = Data} = State,
  ...
```
because it's immediately clear that there no pattern matching on the State variable.

PR #243, opened 2020-04-18
Clean up targeted backwards compatibility

xspirus:
This commit cleans up the targeted implementation from some unneeded
backwards compatible macros. This should also improve the performance
of the targeted testing.


src/proper.erl
```diff
@@ -1388,27 +1389,29 @@ run(Result, #ctx{mode = Mode, bound = Bound} = Ctx, _Opts) when is_boolean(Resul
...
 	    {error, too_many_instances}
     end;
-run({exists, TMap, Prop, Not}, #ctx{mode = new} = Ctx,
+run({exists, RawType, Prop, Not}, #ctx{mode = new} = Ctx,
     #opts{search_strategy = Strat, search_steps = Steps,
           output_fun = Print, start_size = StartSize} = Opts) ->
-    proper_target:init_strategy(Strat),
-    Target = proper_target:targeted(TMap),
-    Print("[", []),
+    proper_target:init_strategy(#{search_strategy => Strat,
```
kostis, line 1395: Since you are changing this code, this is better written in two lines as:
```erlang
    InitOpts = #{search_strategy => Strat, search_steps => Steps}),
    proper_target:init_strategy(InitOpts),
```

src/proper_sa.erl
```diff
@@ -109,8 +108,8 @@
 %% number of the search steps and the strategy.
 
 %% @private
--spec init_strategy(proper:setup_opts()) -> sa_data().
-init_strategy(#{numtests := Steps}) ->
+-spec init_strategy(pos_integer()) -> sa_data().
```
kostis, line 111: Better introduce a type for this (in some target-related file and use throughout)
```erlang
-type search_steps() :: pos_integer().
```

kostis:
Other than the two small comment, looks good overall.

xspirus:
@kostis I have updated the code with the recommendations.

PR #245, opened 2020-04-23
Stateful Targeted Property-Based Testing

xspirus:
This pull request aims to add a TPBT implementation for stateful systems (both `proper_statem` and `proper_fsm`). This is achieved by adding new `targeted_commands/1` and `targeted_commands/2` in both of those. These new commands should be used in conjuction with the `?FORALL_TARGETED` macro.

Examples can be found in the `examples/car_statem.erl` and `examples/car_fsm.erl` files. Both implement the same example, one as a `gen_server -> proper_statem` and one as a `gen_statem -> proper_fsm`.

The examples are also now included in PropEr's test suite, in order to check that everything works as expected. However, tests should be revisited in order to add some more unit testing in the functions added in `proper_statem` and `proper_fsm` respectively.

### DIFF
--- a/examples/car_fsm.erl
+++ b/examples/car_fsm.erl
@@ -52,8 +52,8 @@
 %% -----------------------------------------------------------------------------
 
 
--define(DISTANCE, 1000).
--define(CONSUMPTION, 10).
+-define(DISTANCE, 800).
+-define(CONSUMPTION, 12).
 -define(SLOW_THRESHOLD, 50).
 -define(NORMAL_THRESHOLD, 100).
 -define(FAST_THRESHOLD, 150).
@@ -342,8 +342,6 @@ postcondition(_, _, S, _, {D, B}) ->
   Wanted = Distance + D >= ?DISTANCE andalso Consumption =< ?CONSUMPTION,
   D >= 0.0 andalso B >= 0.0 andalso not Wanted.
 
-weight(_, _, ?CALL(brake, _)) -> 2;
-weight(_, _, ?CALL(travel, _)) -> 4;
 weight(_, _, _) -> 1.
 
 
@@ -396,7 +394,7 @@ prop_distance() ->
 %% provides failing command sequencies more consistently.
 prop_distance_targeted() ->
   ?FORALL_TARGETED(
-     Cmds, more_commands(3, proper_fsm:targeted_commands(?MODULE)),
+     Cmds, proper_fsm:targeted_commands(?MODULE),
      ?TRAPEXIT(
         begin
           start_link(),
@@ -414,7 +412,7 @@ prop_distance_targeted() ->
 prop_distance_targeted_init() ->
   State = {initial_state(), initial_state_data()},
   ?FORALL_TARGETED(
-     Cmds, more_commands(3, proper_fsm:targeted_commands(?MODULE, State)),
+     Cmds, proper_fsm:targeted_commands(?MODULE, State),
      ?TRAPEXIT(
         begin
           start_link(),
--- a/examples/car_fsm.erl
+++ b/examples/car_fsm.erl
@@ -39,12 +39,12 @@
 %% API
 -export([start_link/0, stop/0, accelerate/1, brake/1, travel/1, refuel/1]).
 %% gen_statem callbacks
--export([init/1, callback_mode/0, stopped/3, slow/3, normal/3, fast/3,
-         too_fast/3, terminate/3, code_change/4]).
+-export([init/1, callback_mode/0, stopped/3, slow/3, fast/3, terminate/3,
+         code_change/4]).
 %% proper_fsm callbacks
 -export([initial_state/0, initial_state_data/0, next_state_data/5,
-         precondition/4, postcondition/5, weight/3]).
--export([stopped/1, slow/1, normal/1, fast/1, too_fast/1]).
+         precondition/4, postcondition/5]).
+-export([stopped/1, slow/1, fast/1]).
 
 
 %% -----------------------------------------------------------------------------
@@ -52,11 +52,8 @@
 %% -----------------------------------------------------------------------------
 
 
--define(DISTANCE, 800).
--define(CONSUMPTION, 12).
--define(SLOW_THRESHOLD, 50).
--define(NORMAL_THRESHOLD, 100).
--define(FAST_THRESHOLD, 150).
+-define(DISTANCE, 1000).
+-define(SLOW_THRESHOLD, 100).
 
 
 %% -----------------------------------------------------------------------------
@@ -132,15 +129,6 @@ slow({call, From}, {travel, Value}, S) ->
 slow({call, From}, {refuel, Value}, S) ->
   refuel_helper(From, Value, S).
 
-normal({call, From}, {accelerate, Value}, S) ->
-  accelerate_helper(From, Value, S);
-normal({call, From}, {brake, Value}, S) ->
-  accelerate_helper(From, -Value, S);
-normal({call, From}, {travel, Value}, S) ->
-  travel_helper(From, Value, S);
-normal({call, From}, {refuel, Value}, S) ->
-  refuel_helper(From, Value, S).
-
 fast({call, From}, {accelerate, Value}, S) ->
   accelerate_helper(From, Value, S);
 fast({call, From}, {brake, Value}, S) ->
@@ -148,15 +136,6 @@ fast({call, From}, {brake, Value}, S) ->
 fast({call, From}, {travel, Value}, S) ->
   travel_helper(From, Value, S);
 fast({call, From}, {refuel, Value}, S) ->
-  refuel_helper(From, Value, S).
-
-too_fast({call, From}, {accelerate, Value}, S) ->
-  accelerate_helper(From, Value, S);
-too_fast({call, From}, {brake, Value}, S) ->
-  accelerate_helper(From, -Value, S);
-too_fast({call, From}, {travel, Value}, S) ->
-  travel_helper(From, Value, S);
-too_fast({call, From}, {refuel, Value}, S) ->
   refuel_helper(From, Value, S).
 
 terminate(_Reason, _StateName, _State) ->
@@ -204,23 +183,15 @@ refuel_helper(From, Value, S) ->
 
 accelerator(Speed, slow) ->
   integer(1, ?SLOW_THRESHOLD - Speed);
-accelerator(Speed, normal) ->
-  integer(max(1, ?SLOW_THRESHOLD - Speed + 1), ?NORMAL_THRESHOLD - Speed);
 accelerator(Speed, fast) ->
-  integer(max(1, ?NORMAL_THRESHOLD - Speed + 1), ?FAST_THRESHOLD - Speed);
-accelerator(Speed, too_fast) ->
-  integer(max(1, ?FAST_THRESHOLD - Speed + 1), ?MAX_SPEED - Speed).
+  integer(max(1, ?SLOW_THRESHOLD - Speed + 1), ?MAX_SPEED - Speed).
 
 braker(Speed, stopped) ->
   exactly(Speed);
 braker(Speed, slow) ->
   integer(max(1, Speed - ?SLOW_THRESHOLD + 1), Speed - 1);
-braker(Speed, normal) ->
-  integer(max(1, Speed - ?NORMAL_THRESHOLD + 1), Speed - ?SLOW_THRESHOLD - 1);
 braker(Speed, fast) ->
-  integer(max(1, Speed - ?FAST_THRESHOLD + 1), Speed - ?NORMAL_THRESHOLD - 1);
-braker(Speed, too_fast) ->
-  integer(1, Speed - ?FAST_THRESHOLD - 1).
+  integer(1, Speed - ?SLOW_THRESHOLD - 1).
 
 traveler() -> integer(1, 100).
 
@@ -253,21 +224,7 @@ slow(S) ->
     [{history, ?CALL(travel, [traveler()])},
      {stopped, ?CALL(refuel, [refueler(Fuel)])}].
 
-normal(S) ->
-  #test_state{fuel = Fuel, speed = Speed} = S,
-  accelerate_commands(Speed) ++
-    brake_commands(Speed) ++
-    [{history, ?CALL(travel, [traveler()])},
-     {stopped, ?CALL(refuel, [refueler(Fuel)])}].
-
 fast(S) ->
-  #test_state{fuel = Fuel, speed = Speed} = S,
-  accelerate_commands(Speed) ++
-    brake_commands(Speed) ++
-    [{history, ?CALL(travel, [traveler()])},
-     {stopped, ?CALL(refuel, [refueler(Fuel)])}].
-
-too_fast(S) ->
   #test_state{fuel = Fuel, speed = Speed} = S,
   accelerate_commands(Speed) ++
     brake_commands(Speed) ++
@@ -336,13 +293,11 @@ next_state_data(_, _, S, _, ?CALL(refuel, [Value])) ->
                distance = Distance + Travelled,
                burnt = B + Burnt}.
 
-postcondition(_, _, S, _, {D, B}) ->
-  #test_state{distance = Distance, burnt = Burnt} = S,
-  Consumption = calculate_consumption(Distance + D, Burnt + B),
-  Wanted = Distance + D >= ?DISTANCE andalso Consumption =< ?CONSUMPTION,
-  D >= 0.0 andalso B >= 0.0 andalso not Wanted.
-
-weight(_, _, _) -> 1.
+postcondition(_, _, S, _, {D, B}) when D >= 0, B >= 0 ->
+  #test_state{distance = Distance} = S,
+  Distance + D < ?DISTANCE;
+postcondition(_, _, _S, _, _R) ->
+  false.
 
 
 %% -----------------------------------------------------------------------------
@@ -353,22 +308,14 @@ weight(_, _, _) -> 1.
 accelerate_commands(Speed) ->
   [{slow, ?CALL(accelerate, [accelerator(Speed, slow)])}
    || Speed < ?SLOW_THRESHOLD] ++
-    [{normal, ?CALL(accelerate, [accelerator(Speed, normal)])}
-     || Speed < ?NORMAL_THRESHOLD] ++
     [{fast, ?CALL(accelerate, [accelerator(Speed, fast)])}
-     || Speed < ?FAST_THRESHOLD] ++
-    [{too_fast, ?CALL(accelerate, [accelerator(Speed, too_fast)])}
      || Speed < ?MAX_SPEED].
 
 brake_commands(Speed) ->
   [{stopped, ?CALL(brake, [braker(Speed, stopped)])}] ++
     [{slow, ?CALL(brake, [braker(Speed, slow)])} || Speed > 1] ++
-    [{normal, ?CALL(brake, [braker(Speed, normal)])}
-     || Speed > ?SLOW_THRESHOLD + 1] ++
     [{fast, ?CALL(brake, [braker(Speed, fast)])}
-     || Speed > ?NORMAL_THRESHOLD + 1] ++
-    [{too_fast, ?CALL(brake, [braker(Speed, too_fast)])}
-     || Speed > ?FAST_THRESHOLD + 1].
+     || Speed > ?SLOW_THRESHOLD + 1].
 
 
 %% -----------------------------------------------------------------------------
@@ -378,23 +325,24 @@ brake_commands(Speed) ->
 
 %% Vanilla property based testing. This should not fail consistently.
 prop_distance() ->
-  ?FORALL(Cmds, proper_fsm:commands(?MODULE),
-          ?TRAPEXIT(
-             begin
-               start_link(),
-               {H, {_, S}, R} = proper_fsm:run_commands(?MODULE, Cmds),
-               stop(),
-               ?WHENFAIL(
-                  on_failure(H, S, R),
-                  aggregate(zip(proper_fsm:state_names(H), command_names(Cmds)),
-                            R =:= ok))
-             end)).
+  ?FORALL(
+     Cmds, more_commands(2, proper_fsm:commands(?MODULE)),
+     ?TRAPEXIT(
+        begin
+          start_link(),
+          {H, {_, S}, R} = proper_fsm:run_commands(?MODULE, Cmds),
+          stop(),
+          ?WHENFAIL(
+             on_failure(H, S, R),
+             aggregate(zip(proper_fsm:state_names(H), command_names(Cmds)),
+                       R =:= ok))
+        end)).
 
 %% Targeted property based testing, where maximizing the distance travelled
 %% provides failing command sequencies more consistently.
 prop_distance_targeted() ->
   ?FORALL_TARGETED(
-     Cmds, proper_fsm:targeted_commands(?MODULE),
+     Cmds, more_commands(2, proper_fsm:targeted_commands(?MODULE)),
      ?TRAPEXIT(
         begin
           start_link(),
@@ -412,7 +360,7 @@ prop_distance_targeted() ->
 prop_distance_targeted_init() ->
   State = {initial_state(), initial_state_data()},
   ?FORALL_TARGETED(
-     Cmds, proper_fsm:targeted_commands(?MODULE, State),
+     Cmds, more_commands(2, proper_fsm:targeted_commands(?MODULE, State)),
      ?TRAPEXIT(
         begin
           start_link(),
@@ -435,9 +383,7 @@ prop_distance_targeted_init() ->
 %% Get state names based on speed.
 state_name(0) -> stopped;
 state_name(S) when S =< ?SLOW_THRESHOLD -> slow;
-state_name(S) when S =< ?NORMAL_THRESHOLD -> normal;
-state_name(S) when S =< ?FAST_THRESHOLD -> fast;
-state_name(_S) -> too_fast.
+state_name(_S) -> fast.
 
 %% Function to be called when the property fails.
 on_failure(History, State, {postcondition, false}) ->
--- a/examples/car_fsm.erl
+++ b/examples/car_fsm.erl
@@ -1,7 +1,7 @@
 %%% -*- coding: utf-8 -*-
 %%% -*- erlang-indent-level: 2 -*-
 %%% -------------------------------------------------------------------
-%%% Copyright 2010-2020 Manolis Papadakis <manopapad@gmail.com>,
+%%% Copyright 2020-     Manolis Papadakis <manopapad@gmail.com>,
 %%%                     Eirini Arvaniti <eirinibob@gmail.com>
 %%%                 and Kostis Sagonas <kostis@cs.ntua.gr>
 %%%
@@ -20,7 +20,7 @@
 %%% You should have received a copy of the GNU General Public License
 %%% along with PropEr.  If not, see <http://www.gnu.org/licenses/>.
 
-%%% @copyright 2010-2020 Manolis Papadakis, Eirini Arvaniti and Kostis Sagonas
+%%% @copyright 2020 Spiros Dontas and Kostis Sagonas
 %%% @version {@version}
 %%% @author Spiros Dontas
 
@@ -29,8 +29,6 @@
 -behaviour(proper_fsm).
 
 -include_lib("proper/include/proper.hrl").
-
--compile([export_all, nowarn_export_all]).
 
 
 %% -----------------------------------------------------------------------------
@@ -47,9 +45,6 @@
 -export([initial_state/0, initial_state_data/0, next_state_data/5,
          precondition/4, postcondition/5, weight/3]).
 -export([stopped/1, slow/1, normal/1, fast/1, too_fast/1]).
-%% Properties
--export([prop_distance/0, prop_distance_targeted/0,
-         prop_distance_targeted_init/0]).
 
 
 %% -----------------------------------------------------------------------------
@@ -164,12 +159,6 @@ too_fast({call, From}, {travel, Value}, S) ->
 too_fast({call, From}, {refuel, Value}, S) ->
   refuel_helper(From, Value, S).
 
-handle_call(From, stop, Data) ->
-  {stop_and_reply, normal, {reply, From, ok}, Data}.
-
-handle_info(Info, StateName, Data) ->
-  {stop, {shutdown, {unexpected, Info, StateName}}, StateName, Data}.
-
 terminate(_Reason, _StateName, _State) ->
   ok.
 
@@ -260,29 +249,29 @@ slow(S) ->
   #state{fuel = Fuel, speed = Speed} = S,
   accelerate_commands(Speed) ++
     brake_commands(Speed) ++
-    [{history, ?CALL(travel, [traveler()])}] ++
-    [{stopped, ?CALL(refuel, [refueler(Fuel)])}].
+    [{history, ?CALL(travel, [traveler()])},
+     {stopped, ?CALL(refuel, [refueler(Fuel)])}].
 
 normal(S) ->
   #state{fuel = Fuel, speed = Speed} = S,
   accelerate_commands(Speed) ++
     brake_commands(Speed) ++
-    [{history, ?CALL(travel, [traveler()])}] ++
-    [{stopped, ?CALL(refuel, [refueler(Fuel)])}].
+    [{history, ?CALL(travel, [traveler()])},
+     {stopped, ?CALL(refuel, [refueler(Fuel)])}].
 
 fast(S) ->
   #state{fuel = Fuel, speed = Speed} = S,
   accelerate_commands(Speed) ++
     brake_commands(Speed) ++
-    [{history, ?CALL(travel, [traveler()])}] ++
-    [{stopped, ?CALL(refuel, [refueler(Fuel)])}].
+    [{history, ?CALL(travel, [traveler()])},
+     {stopped, ?CALL(refuel, [refueler(Fuel)])}].
 
 too_fast(S) ->
   #state{fuel = Fuel, speed = Speed} = S,
   accelerate_commands(Speed) ++
     brake_commands(Speed) ++
-    [{history, ?CALL(travel, [traveler()])}] ++
-    [{stopped, ?CALL(refuel, [refueler(Fuel)])}].
+    [{history, ?CALL(travel, [traveler()])},
+     {stopped, ?CALL(refuel, [refueler(Fuel)])}].
 
 precondition(stopped, stopped, _S, {call, _, refuel, _}) ->
   true;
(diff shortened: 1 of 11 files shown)


src/proper_fsm.erl
```diff
@@ -233,6 +238,33 @@ commands(Mod, {Name,Data} = InitialState) ->
 	 proper_statem:commands(?MODULE, State),
 	 [{init,InitialState}|Cmds]).
 
+%% @doc A special PropEr type which generates targeted command sequences,
+%% according to a finite state machine specification and taking into
+%% consideration a utility value. The functiona takes as input the name of
```
kostis, line 243: typo here

src/proper_fsm.erl
```diff
@@ -313,6 +345,119 @@ postcondition(#state{name = From, data = Data, mod = Mod}, Call, Res) ->
...
+%% @private
+-spec targeted_commands_gen(mod_name(), fsm_state()) -> proper_types:type().
+targeted_commands_gen(Mod, {Name, Data} = InitialState) ->
+  State = #state{name = Name, data = Data, mod = Mod},
+  ?LET([_ | Cmds], proper_statem:commands(?MODULE, State),
+       {finalize_weights(State, Cmds, maps:new()),
+        [{init, InitialState} | Cmds]}).
+
+%% @private
+-spec next_commands_gen(mod_name()) -> fun((_, _) -> proper_types:type()).
```
kostis, line 375: Perhaps make `fun((_, _) -> proper_types:type())` a type?  (since it is also used in other places)
Also, are the `_`s really `any()` ?

src/proper_fsm.erl
```diff
@@ -313,6 +345,119 @@ postcondition(#state{name = From, data = Data, mod = Mod}, Call, Res) ->
...
+  ?LET(Transitions,
+       ?LAZY(get_transitions(Mod, From, Data)),
+       begin
+         ValidTransitions = [{To, Cmd} || {To, Cmd} <- Transitions,
+                                          precondition(State, Cmd)],
+         WeightedCmds = lists:map(SelectAux, ValidTransitions),
+         proper_types:frequency(WeightedCmds)
+       end).
+
+%% Track all the state transitions and add these to the weights map.
```
kostis, line 434: these -> them in the comment

src/proper_gen_next.erl
```diff
@@ -203,6 +204,22 @@ apply_temperature_scaling(Generator) ->
       end
   end.
 
+%% Add any parameters that were added to the original generator,
+%% so that they apply in the targeted generation as well.
+
+apply_parameters(Gen, RawType) ->
```
kostis, line 210: suggestion:
  factor out the repeated construction of the `?LAZY(Gen(Base, Temp)))` in this code?  (i.e. assign it in a variable before the `Type = ... ` line?

src/proper_statem.erl
```diff
@@ -294,6 +302,12 @@
 -callback next_state(symbolic_state() | dynamic_state(), term(),
 		     symbolic_call()) -> symbolic_state() | dynamic_state().
 
+-callback all_commands(symbolic_state()) -> [symbolic_call()].
```
kostis, line 305: Totally minor comment here, but it feels inconsistent to list the two callbacks in one order and in the exact opposite in the list of optional_callbacks.

src/proper_statem.erl
```diff
@@ -462,6 +476,191 @@ move_shrinker(Seq, Par, I) ->
...
+  maps:map(fun (_Key, W) ->
+               Random = rand:uniform(1 + (Max - Min)) - (1 - Min),
+               NW = W + Random,
+               case NW >= 1 of
+                 true -> NW;
+                 false -> 1
+               end
+           end, Weights).
+
+%% Used to return the generator the commands of the targeted
```
kostis, line 566: Phrasing of this comment is unparseable.

src/proper_statem.erl
```diff
@@ -462,6 +476,191 @@ move_shrinker(Seq, Par, I) ->
...
+          Cmds,
+          ?LET(List,
+               ?SIZED(Size,
+                      proper_types:noshrink(
+                        next_gen(Size * ?RESIZE_FACTOR, Mod,
+                                 InitialState, 1, Weights))),
+               proper_types:shrink_list(List)),
+          is_valid(Mod, InitialState, Cmds, []))).
+
+%% Used to return the generator the commands of the targeted
```
kostis, line 583: Same comment as above

src/proper_statem.erl
```diff
@@ -462,6 +476,191 @@ move_shrinker(Seq, Par, I) ->
...
+%% generation with an initial state provided by the user.
+
+%% @private
+-spec next_gen(mod_name(), symbolic_state(), weights()) -> proper_types:type().
```
kostis, line 587: Comment about `next_gen/2` and `next_gen/3`:
The convention/tradition in Erlang is that when there exist such functions the _additional_ arguments appear last in the function with arity+1.  Is this an issue here?

src/proper_statem.erl
```diff
@@ -980,3 +1179,7 @@ spawn_link_cp(ActualFun) ->
 	      ActualFun()
 	  end,
     spawn_link(Fun).
+
+-spec is_exported(mod_name(), {fun_name(),arity()}) -> boolean().
+is_exported(Mod, Fun) ->
```
kostis, line 1184: `Fun` -> `FA` (or similar name -- using `Fun` is confusing)

src/proper_target.erl
```diff
@@ -248,6 +281,11 @@ strategy(Strat) ->
       Strat
   end.
 
+%% @private
+get_stateful_cmds({'$used', Used, {Weights, Cmds}})
+  when is_map(Weights) andalso is_list(Cmds) ->
```
kostis, line 286: `,` instead of `andalso` here suffices

src/proper_target.erl
```diff
@@ -265,20 +303,35 @@ init([{Strategy, Data}]) ->
...
 handle_call({init_target, Type}, _From, State) ->
   #state{strategy = Strat} = State,
   NextFun = proper_gen_next:from_proper_generator(Type),
   NewTarget = Strat:init_target(Type, NextFun),
   {reply, ok, State#state{target = NewTarget}};
+
+handle_call(init_stateful, _From, State) ->
+  {reply, ok, State#state{stateful = true}};
+
+handle_call({pdict, Values}, _From, State) ->
```
kostis, line 326: `KVs` or `KeyValues` instead of `Values` here.

test/proper_tests.erl
```diff
@@ -1366,8 +1366,11 @@ sampleshrink_test_() ->
 %%------------------------------------------------------------------------------
 
 examples_are_ok_test_() ->
-    [{timeout, 42, ?_assertEqual([], proper:module(M))}
-     || M <- [b64,elevator_fsm,ets_statem,mastermind,pdict_statem,stack]].
+    %% No need to test more in order to check that the examples work.
```
kostis, line 1369: This change is not OK.  I suggest we keep running as many tests as we can with the default `numtests` and we only run the two car tests with 42.

test/proper_tests.erl
```diff
@@ -1366,6 +1366,7 @@ sampleshrink_test_() ->
 %%------------------------------------------------------------------------------
 
 examples_are_ok_test_() ->
+    %% No need to test more in order to check that the examples work.
```
kostis, line 1369: This comment does not belong here -- it's leftover from before the changes.

examples/car_fsm.erl
```diff
@@ -0,0 +1,543 @@
...
+%% -----------------------------------------------------------------------------
+
+
+accelerate_helper(From, Value, S) ->
+  #gen_state{fuel = Fuel, speed = Speed} = S,
+  {Distance, Acceleration, Burnt} =
+    acceleration_calculations({Speed, Value}, Fuel),
+  StateName = state_name(Speed + Acceleration),
+  gen_statem:reply(From, {Distance, Burnt}),
+  {next_state, StateName, S#gen_state{fuel = Fuel - Burnt,
```
kostis, line 180: I think these two lines look better as
```erl
  NS = S#gen_state{fuel = Fuel - Burnt, speed = Speed + Acceleration},
  {next_state, StateName, NS}.
```
(similarly in other places in this module).

examples/car_fsm.erl
```diff
@@ -0,0 +1,543 @@
...
+-define(HOUR, 3600).
+-define(ACCELERATION, 5).
+-define(DECELERATION, 20).
+-define(MAX_FUEL, 70).
+-define(MAX_SPEED, 200).
+-define(AVG(X), lists:sum(X) / length(X)).
+-define(SLOW_THRESHOLD, 50).
+-define(NORMAL_THRESHOLD, 100).
+-define(FAST_THRESHOLD, 150).
+-define(CALL(C, A), {call, ?MODULE, C, A}).
```
kostis, line 66: The PropEr(TM) use of this `?CALL` macro is unclear.  The module uses it at some places, but not in other cases that it would look reasonable to also use it (e.g. in `weight`, `precondition`, `postcondition`, etc.) it does not.  Either use it consistently, or take it out.

examples/car_fsm.erl
```diff
@@ -0,0 +1,543 @@
...
+    [{too_fast, ?CALL(brake, [braker(Speed, too_fast)])}
+     || Speed > ?FAST_THRESHOLD + 1].
+
+
+%% -----------------------------------------------------------------------------
+%% Properties
+%% -----------------------------------------------------------------------------
+
+
+%% This should "never" fail:<br/>
```
kostis, line 379: This comment is misleading.  "Never" is too strong a word here.  "Rarely" is a better one.
But the intention of having this comment here (and similar ones below) is unclrear.

examples/car_fsm.erl
```diff
@@ -0,0 +1,543 @@
...
+        begin
+          start_link(),
+          {_H, {_, S}, R} = proper_fsm:run_commands(?MODULE, Cmds),
+          stop(),
+          #state{distance = Distance, burnt = Burnt} = S,
+          Consumption = case Distance > 0 of
+                          true -> 100 * Burnt / Distance;
+                          false -> 0
+                        end,
+          UV = Distance,
```
kostis, line 416: No need for the `UV` variable, I think.

examples/car_fsm.erl
```diff
@@ -0,0 +1,543 @@
...
+%% `proper:quickcheck(car_fsm:prop_normal_distance(), 1000).'
+prop_distance() ->
+  ?FORALL(Cmds, proper_fsm:commands(?MODULE),
+          ?TRAPEXIT(
+             begin
+               start_link(),
+               {_H, {_, S}, R} = proper_fsm:run_commands(?MODULE, Cmds),
+               stop(),
+               #state{distance = Distance, burnt = Burnt} = S,
+               Consumption = case Distance > 0 of
```
kostis, line 389: The
```erlang
case Distance > 0 of
   ...
```
part gets repeated 3 times in these properties.
Perhaps it should be abstracted to a function (`calculate_consumption`?)
Should that function also include the previous line and take `S` as argument returning `{Distance, Consumption}` ?

examples/car_fsm.erl
```diff
@@ -0,0 +1,543 @@
...
+          ?MAXIMIZE(UV),
+          ?WHENFAIL(
+             io:format("Distance: ~p~nConsumption: ~p~n",
+                       [Distance, Consumption]),
+             aggregate(command_names(Cmds),
+                       R =:= ok andalso (Distance < ?DISTANCE orelse
+                                         Consumption > ?CONSUMPTION)))
+        end)).
+
+%% This should fail most of the times with:<br/>
```
kostis, line 426: Instead of this comment, better write here how this property differs from the one above and what its purpose is.

examples/car_fsm.erl
```diff
@@ -0,0 +1,543 @@
...
+  Consumption = fuel_consumption(Speed),
+  Burn = Consumption * Distance / 100,
+  case Burn > Fuel of
+    true -> {Fuel * 100 / Consumption, Fuel};
+    false -> {Distance, Burn}
+  end;
+travel_calculations(_D, _S, _F) ->
+  {0, 0.0}.
+
+acceleration_calculations({Speed, Accel}, Fuel) when Accel > 0 ->
```
kostis, line 467: Wouldn't it be better if this function takes its first argument (a 2-tuple) as two separate arguments instead?
It seems unnecessary to wrap them and unwrap them all the time...

examples/car_fsm.erl
```diff
@@ -0,0 +1,543 @@
...
+  Acceleration = case Speed + Accel > ?MAX_SPEED of
+                   true -> ?MAX_SPEED - Speed;
+                   false -> Accel
+                 end,
+  Consumption = fuel_consumption(Speed, Acceleration),
+  Distance = calculate_distance(Speed, Acceleration),
+  Burn = Consumption * Distance / 100,
+  case Burn > Fuel of
+    true -> acceleration_calculations({Speed, Acceleration - ?ACCELERATION},
```
kostis, line 476: I find this two line combination to be of terrible style!
Looks _much_ better as
```erlang
  true ->
    acceleration_calculations(...)
```

src/proper_fsm.erl
```diff
@@ -153,8 +153,11 @@
 
 -export([commands/1, commands/2, run_commands/2, run_commands/3,
 	 state_names/1]).
--export([command/1, precondition/2, next_state/3, postcondition/3]).
+-export([targeted_commands/1, targeted_commands/2]).
+-export([command/1, precondition/2, next_state/3, postcondition/3,
+         all_commands/1]).
```
kostis, line 158: I think it's better that `all_commands/1` appears in the same `-export(..)` as `select_commands/2` below.
Similarly for all other functions that are PropEr-private, and an appropriate comment is placed above that export declaration that these functions are not part of the API.

src/proper_gen_next.erl
```diff
@@ -60,7 +60,9 @@
 -define(GEN_NEXT_DEPTH,  proper_gen_next_cache_depth).
 
 -type temperature() :: float().
--type nf() :: fun((pos_integer(), temperature()) -> term()).
+-type depth() :: pos_integer().
+-type nf_temp() :: temperature() | {depth(), temperature()} | null.
+-type nf() :: fun((any(), nf_temp()) -> proper_types:type()).
```
kostis, line 65: The `nf()` fun changed type for its first argument from `pos_integer()` to `any()`.
1. Is this correct?  Was it wrong before?
2. If it indeed can accept any term in its first arg, the type should read `term()` there as in the type below.

src/proper_statem.erl
```diff
@@ -462,6 +480,189 @@ move_shrinker(Seq, Par, I) ->
...
+                             proper_types:noshrink(
+                               next_gen(Size * ?RESIZE_FACTOR, Mod,
+                                        InitialState, 1, Weights))),
+                      proper_types:shrink_list(List)),
+                 [{init, InitialState} | CmdTail]),
+            is_valid(Mod, InitialState, Cmds, [])).
+
+-spec next_gen(proper_gen:size(), mod_name(), symbolic_state(), pos_integer(),
+               weights()) -> proper_types:type().
+next_gen(Size, Mod, State, Count, Weights) ->
```
kostis, line 607: My previous comment about the various `next_gen` functions being extensions of the previous one and the Erlang argument convention is applicable here too.  It's better as:
```erlang
next_gen(Mod, Weights, State, Mod, Count) ->
```

examples/car_statem.erl
```diff
@@ -0,0 +1,416 @@
...
+%%% @author Spiros Dontas
+
+-module(car_statem).
+-behaviour(gen_server).
+-behaviour(proper_statem).
+
+-include_lib("proper/include/proper.hrl").
+
+
+%% -----------------------------------------------------------------------------
```
kostis, line 34: The relative order of exports and defines differs in this file from that of `car_fsm`.
They should be made consistent (and the `car_fsm` is better).

examples/car_statem.erl
```diff
@@ -0,0 +1,416 @@
...
+
+
+init([]) ->
+  {ok, #gen_state{fuel = ?MAX_FUEL, speed = 0}}.
+
+handle_call({accelerate, Value}, _From, S) ->
+  #gen_state{fuel = Fuel, speed = Speed} = S,
+  {Distance, Acceleration, Burnt} =
+    acceleration_calculations({Speed, Value}, Fuel),
+  {reply, {Distance, Burnt}, S#gen_state{fuel = Fuel - Burnt,
```
kostis, line 117: Similar stylistic comment as `car_fsm` here.

examples/car_statem.erl
```diff
@@ -0,0 +1,416 @@
...
+weight(travel) -> 4;
+weight(_) -> 1.
+
+
+%% -----------------------------------------------------------------------------
+%% Properties
+%% -----------------------------------------------------------------------------
+
+
+%% This should "never" fail:
```
kostis, line 258: Same comment

examples/car_statem.erl
```diff
@@ -0,0 +1,416 @@
...
+%% `proper:quickcheck(car_statem:prop_normal_distance(), 1000).'
+prop_distance() ->
+  ?FORALL(Cmds, commands(?MODULE),
+          ?TRAPEXIT(
+             begin
+               start_link(),
+               {_H, S, R} = run_commands(?MODULE, Cmds),
+               stop(),
+               #state{distance = Distance, burnt = Burnt} = S,
+               Consumption = case Distance > 0 of
```
kostis, line 268: Same comment

examples/car_statem.erl
```diff
@@ -0,0 +1,416 @@
...
+acceleration_calculations({Speed, Accel}, Fuel) when Accel > 0 ->
+  Acceleration = case Speed + Accel > ?MAX_SPEED of
+                   true -> ?MAX_SPEED - Speed;
+                   false -> Accel
+                 end,
+  Consumption = fuel_consumption(Speed, Acceleration),
+  Distance = calculate_distance(Speed, Acceleration),
+  Burn = Consumption * Distance / 100,
+  case Burn > Fuel of
+    true -> acceleration_calculations({Speed, Acceleration - ?ACCELERATION},
```
kostis, line 356: Same comment

kostis:
A (final?) general comment / suggestion about the two examples.

Given the amount of code (in auxiliary functions) that the two example modules share, I am wondering whether it would be better to factor out this code either in a `car.erl` module or even in a `car.inc` auxiliary file which can simply be included in both.
Besides better maintainability, this may allow the reader to concentrate on the _core_ of the examples (statem vs fsm stuff that need to be done) instead of having to look at a lot of "boring" code at the same time.

src/proper_statem.erl
xspirus: This was "inspired" by the similar `commands/4` function, already present in this module. That's why the `Size` was put as the first argument.

src/proper_gen_next.erl
xspirus: 1. Yes this was wrong, as the `nf()` is the neighborhood function provided by the user. It's signature is `fun (Base, Temp) -> ...`, where `Base` depends on the initial `RawType`, so this is not a `pos_integer()` for sure.
2. Yes, will update to `term()`.

src/proper_statem.erl
kostis: Perhaps there a good reason to leave `Size` first, but
1. without a comment in that function, nobody is aware of this reason, and
2. this does not explain the placement of `Weights` last.

xspirus: As long as it's OK with you, I will update my code :)

examples/car_fsm.erl
```diff
@@ -0,0 +1,543 @@
+%%% -*- coding: utf-8 -*-
+%%% -*- erlang-indent-level: 2 -*-
+%%% -------------------------------------------------------------------
+%%% Copyright 2020-     Manolis Papadakis <manopapad@gmail.com>,
```
TheGeorge, line 4: Double copyroght information

examples/car_fsm.erl
```diff
@@ -0,0 +1,543 @@
...
+-define(CALL(C, A), {call, ?MODULE, C, A}).
+-define(NAME, car).
+
+
+%% -----------------------------------------------------------------------------
+%% Records
+%% -----------------------------------------------------------------------------
+
+
+-record(gen_state,
```
TheGeorge, line 75: The naming is a bit confusing here:
gen_<something> typically refers to a generic behaviour. gen_state sounds like some generic implementation for some persistent state.

TheGeorge:
Do you have some higher-level documentation as to how SPBT is implemented?

src/proper_target.erl
```diff
@@ -265,20 +303,35 @@ init([{Strategy, Data}]) ->
...
   #state{strategy = Strat} = State,
   NextFun = proper_gen_next:from_proper_generator(Type),
   NewTarget = Strat:init_target(Type, NextFun),
   {reply, ok, State#state{target = NewTarget}};
+
+handle_call(init_stateful, _From, State) ->
+  {reply, ok, State#state{stateful = true}};
+
+handle_call({pdict, KVs}, _From, State) ->
+  lists:foreach(fun ({K, V}) -> put(K, V) end, KVs),
```
TheGeorge, line 327: List constructions are faster and more expressive here:
[put(K, V) || {K, V} <- KVs]

Also, we are using a gen_server an you still write stuff to the process dictionary. This should really be stored in the gen_server state instead. Please change

src/proper_target.erl
```diff
@@ -265,20 +303,35 @@ init([{Strategy, Data}]) ->
 handle_call(gen, _From, State) ->
   #state{strategy = Strat, target = Target, data = Data} = State,
   {NextValue, NewTarget, NewData} = Strat:next(Target, Data),
-  {reply, NextValue, State#state{target = NewTarget, data = NewData}};
+  Ret = case State#state.stateful of
```
TheGeorge, line 306: Why is this not part of the generator? What makes stateful so different that you need this extra code-path. For non-targeted generators this is not the case and having it for targeted ones is a bit strange.

src/proper_target.erl
```diff
@@ -224,7 +257,7 @@ reset() ->
 safe_call(Pid, Call) ->
```
TheGeorge, line 257: Just saw this, what is this for? calling a gen_server that crashed returns ok?

src/proper_target.erl
```diff
@@ -170,25 +173,55 @@ targeted(RawType) ->
...
 targeted_gen() ->
+  update_pdict(['$left', '$size']),
   TargetserverPid = get('$targetserver_pid'),
   gen_server:call(TargetserverPid, gen).
 
 %% @doc Get the shrinker for a Type.
 
 -spec get_shrinker(proper_types:type()) -> proper_types:type().
 get_shrinker(Type) ->
-  TargetserverPid = get('$targetserver_pid'),
```
TheGeorge, line 190: This is not part of the changes, but I believe this needs to be refactored. Uses of get and put like this are wrong. This should be part of the propers run-state

src/proper_target.erl
kostis: > List constructions are faster and more expressive here:
> [put(K, V) || {K, V} <- KVs]

This is not a good argument.  IMO, the `lists:foreach/2` is the PropEr call here.
(In general, when you want to execute something for its side-effects only and not construct a _resulting_ list.  It makes the intention clear; a comprehension does not.)

xspirus: These values are needed, so that when someone is using `?SIZED` in a `?USERNF` neighborhood function, then the `size` is correctly provided (as the `$size` variable exists and gets modified in the main `proper`'s process dictionary, see [`proper:get_size()`](https://github.com/proper-testing/proper/blob/master/src/proper.erl#L660)).

src/proper_target.erl
xspirus: For stateful targeted property testing, what two values are generated, one containing a `map` with command names as keys and weight as a value, and the other containing the actual commands. The actual search occurs on the weights of each command, instead of directly on the commands generated. However, I did not want to alter the already existing API of stateful commands, that is why internally I need to keep both values, while the API needs only return the command sequence to `proper` and subsequently the user.

src/proper_target.erl
xspirus: The reason this exists, is because at the moment the `?FORALL_TARGETED` and `?EXISTS`-`?NOT_EXISTS` macros work a bit different.
exv
**`?FORALL_TARGETED`**

Essentially, this is a wrapper for a `?FORALL` macro, which also sets up the `proper_target` `gen_server`. That means that the `gen_server` is "up" on all stages of testing (run and shrink).

**`?EXISTS` - `?NOT_EXISTS`**

These use the custom `perform_search` function, however the `proper_target` `gen_server` is only "up" during that phase. When shrinking the `gen_server` is "down", but the property keeps making calls to it through the `?MAXIMIZE`-`?MINIMIZE` macros. These are irrelevant during shrinking however, so a non existent `gen_server` should not raise any exception.

src/proper_target.erl
xspirus: You are correct. However, at the moment the same thing is followed in `proper_typeserver`'s codebase. In order to update this, there would need to be a refactoring to `proper` module in order to work as a `gen_server` probably.

src/proper_target.erl
TheGeorge: I'm mostly arguing against the usage of the process dictionary. It's local to the gen_server process and as such only accessible from within it. They should really be stored in the gen_server state. 

How do you read these values?

src/proper_target.erl
TheGeorge: It seems like this can lead to some relatively hard to detect bugs. E.g. the gen_server crashed somehow and all calls seem to succeed. 

In this case the system should loudly crash. Is there not the possibility(in another PR) to have the mode/phase forwarded to the API and ignore them there if we are shrinking?

src/proper_target.erl
xspirus: They are being used during the call `proper_gen:safe_generate`, which then calls `proper:get_size`. So this is something unrelated to the targeted implementation, but is used by other parts of PropEr and I wanted a way to make sure that `?SIZED` generators worked as expexted when using a user provided NF. Without these updates, the gen_server does not obtain the updated $size, so the generation never grows in size.

src/proper_target.erl
xspirus: Yes, this is something that has been discussed.

src/proper_target.erl
TheGeorge: I think there is a misunderstanding here:
- I don't argue to have or not have the values stored
- the use of the process dictionary (`get/1` and `put/2`) does not make sense as every function that can access the dictionary also gets the `gen_server` state as an argument. So store these values in the state instead.
- If you read these values from outside the `gen_server` process (outside handle_... of this module) you are reading different values.

xspirus: Again, PropEr uses a `$size` variable in its process dictionary which is used whenever a `?SIZED` generator is generating values. Adding `size` to the `gen_server`'s state would not fix the problem, as while the call to `proper_gen:safe_generate` occurs within a `handle_call`, there is no way to let that function know of a new `$size` value produced by `proper:grow_size` calls.

To elaborate a bit further, if we exclude these updates to the `proper_target` `gen_server`'s process dictionary, then that has a value of `$size` equal to the `start_size`, which never gets updated. So if a user provided a neighborhood function with `?SIZED` in it, then the size would never grow.

Again, as mentioned below, all of this could be avoided by refactoring the main `proper` module to a `gen_server`, but this should be part of another PR.

It's not clear to me whether there is a better solution to the above problem for the moment, but I am open to suggestions.

xspirus:
@TheGeorge I will add some high level documentation of how to use the STPBT part of proper in both `proper_statem` as well as `proper_fsm` (with a brief explanation of how the algorithm works).

src/proper_fsm.erl
```diff
@@ -147,14 +147,37 @@
 %%%                    cleanup(),
 %%%                    Result =:= ok
 %%%                end).'''
+%%%
+%%% == Targeted Testing ==
```
kostis, line 151: Add the word "Stateful" here -- also, perhaps in the paragraph below.

src/proper_fsm.erl
```diff
@@ -314,6 +367,108 @@ postcondition(#state{name = From, data = Data, mod = Mod}, Call, Res) ->
...
+
+%% @private
+-spec select_command(state(), weights()) -> proper_types:type().
+select_command(State, Weights) ->
+  #state{name = From, data = Data, mod = Mod} = State,
+  SelectAux =
+    fun ({To, {call, _Mod, Call, _Args} = SymbCall}) ->
+        RealTo = cook_history(From, To),
+        Key = {From, RealTo, Call},
+        case maps:is_key(Key, Weights) of
```
kostis, line 427: I think this `case` is better if it computes/returns the `Weight` element, with the `{Weight, SymbCall}` 2-tuple construction being in its own line below the `case`.

src/proper_fsm.erl
```diff
@@ -314,6 +367,108 @@ postcondition(#state{name = From, data = Data, mod = Mod}, Call, Res) ->
...
+finalize_weights(_S, [], Weights) -> Weights;
+finalize_weights(InitialState, Cmds, Weights) ->
+  FinWeightsAux =
+    fun ({init, _InitialState}, Acc) -> Acc;
+        ({set, Var, {call, _Mod, Call, _Args} = SymbCall}, {State, Ws}) ->
+        NextState = next_state(State, Var, SymbCall),
+        From = State#state.name,
+        To = NextState#state.name,
+        Key = {From, To, Call},
+        case maps:is_key(Key, Ws) of
```
kostis, line 457: Similar comment to the above here about `NewWs` computation and with `{NextState, NewWs}` immediately below it.

src/proper_statem.erl
```diff
@@ -139,6 +139,16 @@
 %%%   function is called both at command generation and command execution time
 %%%   in order to update the model state, therefore the state `S' and the
 %%%   result `Res' can be either symbolic or dynamic.</p></li>
+%%% <li>`all_commands(S::'{@type symbolic_state()}`) ->' {@type [symbolic_call()]}
+%%%   <p>This is an optional callback normally, however it is required for targeted
+%%%   property based testing. Lists all valid symbolic calls, given the current
+%%%   state `S' of the abstract state machine. This list is then filtered based on
+%%%   the preconditions.</p></li>
+%%% <li>`weight(Call::'{@type symbolic_call()}`) -> integer()'
```
kostis, line 147: This reads better as
"When this optional callback is defined and exported, it is used to initialize the weights of commands during targeted property based testing. Otherwise, commands are are given a default initial weight of 1."

src/proper_statem.erl
```diff
@@ -219,18 +229,40 @@
...
+%%%        ?FORALL_TARGETED(Cmds, proper_statem:targeted_commands(?MODULE),
+%%%                         begin
+%%%                             {History, State, Result} = proper_statem:run_commands(?MODULE, Cmds),
+%%%                             UV = uv(History, State, Result),
+%%%                             ?MAXIMIZE(UV),
+%%%                             cleanup(),
+%%%                             Result =:= ok
+%%%                         end).'''
+%%%
+%%% Please note that the `UV' value can be computed in any way fit, depending
```
kostis, line 249: Take out "Please" and `uv' -> `uv/3'

src/proper_statem.erl
```diff
@@ -462,6 +510,189 @@ move_shrinker(Seq, Par, I) ->
...
+  proper_types:add_prop(is_user_nf_stateful, true,
+                        ?USERNF(targeted_commands_gen(Mod),
+                                next_commands_gen(Mod))).
+
+%% @doc Similar to {@link targeted_commands/1}, but generated command sequences
+%% always start at a given state. In this case, the first command is always
+%% `{init, InitialState}' and is used to correctly initialize the state
+%% every time the command sequence is run (i.e. during normal execution,
+%% while shrinking and when checking a counterexample). In this case,
+%% `Mod:initial_state/0' is never called.
```
kostis, line 535: never -> not

src/proper_statem.erl
```diff
@@ -462,6 +510,189 @@ move_shrinker(Seq, Par, I) ->
...
+-spec targeted_commands(mod_name(), symbolic_state()) -> proper_types:type().
+targeted_commands(Mod, InitialState) ->
+  proper_types:add_prop(is_user_nf_stateful, true,
+                        ?USERNF(targeted_commands_gen(Mod, InitialState),
+                                next_commands_gen(Mod, InitialState))).
+
+%% @private
+-spec targeted_commands_gen(mod_name()) -> proper_types:type().
+targeted_commands_gen(Mod) ->
+  ?LET(Cmds,
```
kostis, line 546: In this and the following three functions decide whether you want `?LET` to span three or just two lines. (two is OK IMO).

src/proper_target.erl
TheGeorge: Wait, so the targeted generators are evaluated in the gen_server process and they are accessing the process dictonary for sized because that's how it was done in the main proper prcoess? :-(

xspirus: Exactly, that is why this was needed :/

src/proper_statem.erl
```diff
@@ -463,6 +493,188 @@ move_shrinker(Seq, Par, I) ->
...
+%% @private
+-spec targeted_commands_gen(mod_name(), symbolic_state()) ->
+        proper_types:type().
+targeted_commands_gen(Mod, InitialState) ->
+  ?LET(Cmds, ?LAZY(commands(Mod, InitialState)),
+       {finalize_weights(Mod, InitialState, Cmds, maps:new()), Cmds}).
+
+%% @private
+-spec next_commands_gen(mod_name()) -> next_fun().
+next_commands_gen(Mod) ->
```
kostis, line 542: Isn't this better written as:
```erlang
next_commands_gen(Mod) ->
  next_commands_gen(Mod, Mod:initial_state()).
```
or is there some other difference in the two functions that I am missing?

xspirus: There is a difference (similar to what already exists for the simple `commands/1` and `commands/2`). If not providing an initial state the commands returned are:

```erlang
[{set, {var,1}, {call,mod,some_call,[some_args]}}, ...]
```

If an initial state is provided then the commands returned are:

```erlang
[{init,some_initial_state}, {set, {var,1}, {call,mod,some_call,[some_args]}}, ...]
```

src/proper_fsm.erl
```diff
@@ -147,14 +147,37 @@
 %%%                    cleanup(),
 %%%                    Result =:= ok
 %%%                end).'''
+%%%
+%%% == Stateful Targeted Testing ==
+%%% During testing of the system's behavior, there may be some failing command
+%%% sequencies that the default property based testing does not find with ease,
```
kostis, line 153: "sequencies" -> "sequences"
"default" -> "random"

examples/car_statem.erl
```diff
@@ -0,0 +1,314 @@
...
+  {reply, {Distance, Burnt}, NS}.
+
+handle_cast(_Msg, S) ->
+  {noreply, S}.
+
+handle_info(_Msg, S) ->
+  {noreply, S}.
+
+terminate(_Reason, _S) ->
+  {ok}.
```
kostis, line 147: This reads weird.  A simple `ok` will do.

xspirus: Oops :)

PR #254, opened 2020-11-06
fix float format in plain_stats_printer/3

flowerett:
Hello,
during my work on updating the [elixir library](https://github.com/alfert/propcheck/pull/179) I found an issue in proper:

Seems like for success case (100.00%)  plain_stats_printer/3 will return asterisks instead of a formatted number:

Check:
```erlang
io:fwrite("====== 5.2 can fail =====~n"),
io:fwrite("~5.2f ~n", [1000 * 30 / 70]),
io:fwrite("~5.2f ~n", [100 * 30 / 30]),
io:fwrite("~5.2f ~n", [100 * 18 / 30]),
io:fwrite("~5.2f ~n", [100 * 1 / 30]).
```
```sh
====== 5.2 can fail =====
*****
*****
60.00
 3.33
```

As a fix I extended formatting to 6 digits but 5.1 also will work:
```erlang
io:fwrite("===== 6.2 ok =====~n"),
io:fwrite("~6.2f ~n", [1000 * 30 / 70]),
io:fwrite("~6.2f ~n", [100 * 30 / 30]),
io:fwrite("~6.2f ~n", [100 * 18 / 30]),
io:fwrite("~6.2f ~n", [100 * 1 / 30]),
io:fwrite("===== 5.1 ok =====~n"),
io:fwrite("~5.1f ~n", [1000 * 30 / 70]),
io:fwrite("~5.1f ~n", [100 * 30 / 30]),
io:fwrite("~5.1f ~n", [100 * 18 / 30]),
io:fwrite("~5.1f ~n", [100 * 1 / 30]),
io:fwrite("==================~n").
``` 

```sh
===== 6.2 ok =====
428.57
100.00
 60.00
  3.33
===== 5.1 ok =====
428.6
100.0
 60.0
  3.3
==================
```

Also, it's quite a while now when `proper` lib was released last time.
So I would like to ask if there are any plans to push a new version?

### DIFF
--- a/src/proper.erl
+++ b/src/proper.erl
@@ -2145,7 +2145,12 @@ with_title(Title) ->
 plain_stats_printer(SortedSample, Print, Title) ->
     print_title(Title, Print),
     Total = length(SortedSample),
-    PrFun = fun ({Cmd,Fr}) -> Print("~5.2f\% ~w~n", [100 * Fr / Total,Cmd]) end,
+    PrFun = fun ({Cmd,Fr}) ->
+        case Fr =:= Total of
+          true  -> Print("100.0\% ~w~n", [Cmd]);
+          false -> Print("~5.2f\% ~w~n", [100 * Fr / Total,Cmd])
+        end
+    end,
     lists:foreach(PrFun, process_sorted_sample(SortedSample)).
 
 -spec print_title(title(), output_fun()) -> 'ok'.


kostis:
Thanks for reporting this issue.  Indeed, when value of the percentage reported is 100.00, the `5.2f` format will print it as asterisks, so perhaps using `6.2f` is better.

But, out of curiosity, does this happen in some real use?  Note that this printing happens when PropEr reports, as percentages, the frequencies with which commands were chosen.  So, the range is always between 0.00  and 100.00 -- never e.g., 428.57 as you use in the example above.

Note that 100.00% frequency means that only one command was chosen/used.  I guess it's possible, but is this something that happens in an actual stateful PBT example?  Also note that only one line will be printed in this case, so perhaps it's better to special case this frequency with a hard-coded 100.00% in such a case with code that looks as follows in line 2148:
```erlang
PrFun = fun ({Cmd,Fr}) ->
            case Fr =:= Total of
              true  -> Print("100.0\% ~w~n", [Cmd]);
              false -> Print("~5.2f\% ~w~n", [100 * Fr / Total,Cmd])
            end
        end,
```

I think I prefer this code there instead.  You can either include it in your PR and then I can merge this, or I can push it in a separate commit.  If you decide to modify your PR, please avoid doing whitespace changes to the file (as your PR currently has). 

Let me know what you want me to do.

flowerett:
Thank you for your reply!
I can update the PR, just have a couple of questions.

- I'm curious why do you suppose that additional condition (and case with a hardcoded value) is better than `~6.2f` that will cover all possible cases?  Is it an optimization that will make the printer function faster?

Of course, 428.57 is an artificial and bad example here, but 100% seems to be quite common.
Currently, it appears in [this test](https://github.com/alfert/propcheck/blob/master/test/verify_verbose_elixir_syntax_test.exs#L31) from the elixir library and indeed we have:
```sh
Fr: 100
Total: 100
Cmd: %{test: VerifyVerboseElixirSyntaxTest}
100.00% %{test: VerifyVerboseElixirSyntaxTest}
``` 
in the output. Not sure if this is correct behavior. Are we misusing the `collect` there?


- I agree formatting is subjective. Just mentioned that it's not consistent in this place:
formatting
and can look different if your editor treats spaces and tabs differently. Just wanted to help 🙌 .

Anyway, I'll revert all whitespace changes.

kostis:
> I'm curious why do you suppose that additional condition (and case with a hardcoded value) is better than ~6.2f that will cover all possible cases? Is it an optimization that will make the printer function faster?

This is a good question - and you could not possibly figure out the reason yourself.  It's none of the above reasons and admittedly the reason is not very deep.  It's just that, after changing the print to ~5.2f, I've updated most tutorials (e.g. [this one](https://proper-testing.github.io/tutorials/PropEr_testing_of_generic_servers.html) at its end) to print all frequencies using a  format with five characters.  I do not want to go again over them and add a space in each line :-)


>  Currently, it appears in this test from the elixir library and indeed we have:
> ...  <SNIP>
> in the output. Not sure if this is correct behavior. Are we misusing the collect there?

I am not so versed in Elixir / PropCheck, but I do not see where the categories or commands are defined in this test.  `collect` can of course be called in all cases, but it really makes sense only when you have multiple categories / different commands.  Anyway, as I wrote above, PropEr _should_ be able to print the frequencies properly even when only one category / command exists, so even if the PropCheck test case is misusing `collect` the printing of 100% should be fixed.


The tabs vs spaces issue is something that I do not have a strong opinion about.  However, under both Emacs and Vim, but also on [GitHub](https://github.com/proper-testing/proper/blob/master/src/proper.erl#L2157), the example you point out looks fine to me.  So I am not bothered by the presence of tabs.  If you are seriously bothered, I think it's better to submit another pull request that fixes just that issue (eliminating tabs in all files under `src`) instead of polluting the commit in this PR.

flowerett:
Thank you for the explanations @kostis! Now it's clear and totally makes sense.
I've updated the PR.

PR #260, opened 2021-02-23
Add 16-bit floats to the Erlang abstract code generator

uabboli:
Also respect the erl_anno abstraction.


kostis:
Thanks for the update!

PR #262, opened 2021-03-09
Erlang abstract code cleanup

kostis:


kostis:
Prompted by a comment I noticed in proper_erlang_abstract_code, I've modified the `weighted_union` function and the `frequency()` type to also accept 0 as a value.  This has made that comment obsolete, and the first commit removes that.

@uabboli : Are there some code simplifications that can be done there?

Also, since I edited the file anyway, I've done a pass of reusing some structures to avoid their (unnecessary) re-construction.

uabboli:
No, I don't think so.

PR #263, opened 2021-03-25
Extend PropEr with parallel execution

pablocostass:
This PR extends PropEr with parallel execution. It is based on the work I did for my undergraduate thesis and has some limitations at the moment (e.g., targeted properties). Nonetheless, it should pretty much work (famous last words) and one can find the relevant options to test with parallel PropEr in the documentation of this branch, so please do take a look at it too.

The PR has two extra commits that I considered relevant but might be dropped upon request, those are:

- **Add verbosity to `make test` output** (df7a7d8)
    Adds verbosity when running the tests, as that way it is easier to check at what point parallel PropEr failed in the suite.

- **Add a workflow to test with parallel PropEr** (4c81af9)
    This commit adds a new workflow to be ran with Github Actions that uses 2 workers to run the test suite and ignores failures in the workflow; still, the warnings/failures are recorded by GA, so they will not be lost. This way, no PR would be blocked by some flakyness coming from parallel PropEr.

In short, this PR brings four new options to the table, all related to this new kind of execution:
1. `{numworkers, <Non_negative_number>}` sets the number of workers to use during testing.
2. `pure` (side effect free) or `impure` (with side effects) tells PropEr the kind of property it is going to test; `impure` properties start nodes to isolate the workers from the others and avoid possible test clashes.
3. `{strategy_fun, <Strategy_function>}` overrides the default function that divides the workload among the total of workers.
4. `{stop_nodes, true | false}` tells PropEr whether it should stop the nodes (if any was started) after finishing testing. This one is mainly used for `proper:module/1,2`, but it will probably be useful for some (e.g., build tools).


Makefile
```diff
@@ -61,6 +61,9 @@ endif
 test-examples:
 	$(REBAR3) eunit --dir=examples_test
 
+parallel-test-examples:
```
kostis, line 64: Rename this to `test-parallel` (for uniformity).
Also, add it to the appropriate place in the `.PHONY:` line.

src/proper.erl
```diff
@@ -660,6 +687,33 @@ tests_at_next_size(Size, #opts{numtests = NumTests, start_size = StartSize,
...
+            end;
+        false ->
+            case NumTest == 0 of
+                true -> StartSize;
+                false ->
+                    Diff = (SizesToTest - 1) div (NumTests - 1),
+                    NumTest * Diff + StartSize
+            end
+    end,
+    case Size >= MaxSize of
```
kostis, line 712: Use `min` here instead of four lines.

pablocostass:
@kostis I had forgotten to add the new make target to the Github Action workflow, so I did that after rebasing on top of the latest changes on the main branch.

I see that the Github Action is still failing due to the cache, however this time it is because the current cache is from the main branch and in said branch we do not build the PLT file with `tools` too, so I guess that in 7 days from now on we will have to trigger another run :/

kostis:
I've pushed a small change to the Makefile that subsumes the addition of `tools` to the PLT -- so please rebase/

In order to achieve some progress, perhaps it's a good idea to temporarily disable the `-Wunknown` from the dialyzer call so that testing can continue with the tests.

pablocostass:
Yup, that commit of yours helped @kostis, thanks! I rebased and pushed with the addition of a commit that disables the `unknown` warning, as you suggested. We can drop it in the future whenever we want :)

.github/workflows/ci.yml
```diff
@@ -49,6 +49,16 @@ jobs:
       - name: Test examples
         run: make test-examples
 
+  parallel-tests:
```
kostis, line 52: For uniformity, this should be called either `parallel-examples` or just `parallel`  (the test part is implied).

src/proper.erl
```diff
@@ -284,6 +284,20 @@
...
+%%% <dt>`{numworkers, <Non_negative_number>}'</dt>
+%%% <dd> Specifies the number of workers to spawn when performing the tests (defaults to 0,
+%%% which is PropEr's default behaviour). Each worker gets their own share of the total of number of tests to perform.</dd>
+%%% <dt>`{strategy_fun, <Strategy_function>}'</dt>
+%%% <dd> Overrides the default function used to split the load of tests among the workers.
+%%% It should be of the type {@link strategy_fun()}.</dd>
+%%% <dt>`pure | impure'</dt>
+%%% <dd> Declares the type of the property, as in pure with no side-effects or state,
+%%% and impure with them. <b>Notice</b>: this option will only be taken into account if
+%%% the the number of workers set is greater than 0. In addition, <i>impure</i> properties
```
kostis, line 296: "the the"

src/proper.erl
```diff
@@ -1173,6 +1272,45 @@ inner_test(RawTest, Opts) ->
 	false -> ShortResult
     end.
 
+%% @private
+%% @doc Runs PropEr in parallel mode, through the use of workers to perform the tests.
+%% Under this mode, PropEr will detect if a property is pure or impure based on the passed options.
```
kostis, line 1277: "PropEr will detect if a property is pure or impure based on the passed options."  is not accurate.
"PropEr needs information whether a property is pure or impure, and this information is passed via an option." is a better sentence here, I think.  (Or something similar.)

src/proper.erl
```diff
@@ -1173,6 +1272,45 @@ inner_test(RawTest, Opts) ->
...
+%% On the later case, it will start a node for every worker that will be spawned in order to
+%% avoid test collisions between them.
+-spec parallel_perform(test(), opts()) -> imm_result().
+parallel_perform(Test, #opts{property_type = pure, numtests = NumTests,
+                             numworkers = NumWorkers, strategy_fun = StrategyFun} = Opts) ->
+    TestsPerWorker = StrategyFun(NumTests, NumWorkers),
```
kostis, line 1283: I would suggest to factor out all lines between
```erlang
    TestsPerWorker = StrategyFun(NumTests, NumWorkers),
    ....
    ok = maybe_stop_cover_server([]),
```
into a separate function and use it both for this clause and the next one (by passing it `[]` or `NodeList`).

This will help both for maintainability and for understanding what is common and what is not.

src/proper.erl
```diff
@@ -1273,21 +1418,102 @@ get_rerun_result(#fail{}) ->
...
+
+-spec perform(non_neg_integer(), test(), opts()) -> imm_result() | 'ok'.
 perform(NumTests, Test, Opts) ->
     perform(0, NumTests, ?MAX_TRIES_FACTOR * NumTests, Test, none, none, Opts).
 
+-spec perform(non_neg_integer(), pos_integer(), test(), opts()) -> imm_result() | 'ok'.
+perform(Passed, NumTests, Test, Opts) ->
+    Size = size_at_nth_test(Passed, Opts),
+    put('$size', Size),
+    perform(Passed, NumTests, 3 * ?MAX_TRIES_FACTOR * NumTests, Test, none, none, Opts).
```
kostis, line 1447: You should probably add a comment where this magic number `3` comes from... what is its role.

pablocostass:
@kostis As the commit that had added the `get_all_application_env()` function also fixed some typos, I rebased to change it to only bring in the latter and also pushed some commits to address your comments.

As I am not really happy with how I did the refactoring of the common code that `parallel_perform/2` had, I left that as a different commit. Review it and tell me if you thing something else should be changed.

src/proper.erl
```diff
@@ -1173,6 +1260,60 @@ inner_test(RawTest, Opts) ->
...
+%% @private
+%% @doc Runs PropEr in parallel mode, through the use of workers to perform the tests.
+%% Under this mode, PropEr needs information whether a property is pure or impure,
+%% and this information is passed via an option.
+%% When testing impure properties, PropEr will start a node for every worker that will be
+%% spawned in order to avoid test collisions between them.
+-spec parallel_perform(test(), opts()) -> imm_result().
+parallel_perform(Test, #opts{property_type = pure, numtests = NumTests,
+                             numworkers = NumWorkers, strategy_fun = StrategyFun} = Opts) ->
+    _ = maybe_start_cover_server([]),
```
kostis, line 1294: This line / call is not needed here; it will be done by `spawn_workers_and_get_result`.  (Or am I missing something?)

pablocostass: Oh no, that was totally my mistake, thanks for pointing it out!

pablocostass: Amended the last commit to remove the line.

kostis (approved):

PR #275, opened 2021-05-18
Some cleanups in the parallel code

kostis:
There is also a place where a type error exists - @pablocostass please take a look


pablocostass:
Yep, I don't remember at which point I wrongly mistyped samples, but from line 2202 to 2206 (in your PR) can be removed. We can always concatenate the samples, most of the time we do not have anything in there.

Otherwise, feel free to change the match in L2202 to
```erlang
 {worker_msg, #pass{performed = PassedRcvd, samples = []}, From, Id} ->
```

kostis:
Let us discuss the issue a bit, once you have some time.

PR #284, opened 2021-07-20
Use covertool as a project_plugin, rather than plugin

michalmuskala:
This was resulting in `covertool` being pulled-in as a dependency needlessly in downstream projects in some situations.

Using it as a `project_plugin` is advised by `covertool` [itself](https://github.com/covertool/covertool#rebar3).

More on plugins vs project plugins: https://rebar3.readme.io/docs/using-available-plugins


kostis:
Thanks for your contribution and your interest in PropEr but tere is a similar PR (#283) that beat this one in a race.... 
I am closing this.

PR #285, opened 2021-07-21
update mail unused email address to current one

TheGeorge:
I realized that this address is still referenced here, but not used anymore


kostis:
Since you are changing these files, perhaps it's a good idea to modify the copyright year to be `-2021` in all of them.  Can you do this as part of this PR?

TheGeorge:
Yes, let me update the PR

TheGeorge:
I updated the dates for all touched files

PR #294, opened 2022-05-27
Address deprecation warnings in OTP25

### LINKED ISSUE #293: Substitute uses of `slave` with `peer`
Erlang/OTP 25.0 introduces the following warnings about the `slave` module becoming deprecated in two years:

```
proper.erl:2407:10: Warning: slave:start_link/2 is deprecated and will be removed in OTP 27; use the 'peer' module instead
% 2407|     case slave:start_link(HostName, SlaveName) of

proper.erl:2479:64: Warning: slave:stop/1 is deprecated and will be removed in OTP 27; use the 'peer' module instead
% 2479|             StopFun = fun({Node, {already_running, false}}) -> slave:stop(Node);
```

Thus, I am creating this issue, which obviously is not urgent (esp. since PropEr still supports Erlang/OTP versions where the `peer` module is not present) so that somebody looks into this.  @pablocostass perhaps?

pablocostass:
Address the deprecation of the old `slave` module in OTP 25.x. To do so, two new macros `?START_PEER_NODE/1` and `?STOP_PEER_NODE/1` have been added, which handle the node setup and teardown properly depending on the OTP version PropEr was compiled with.

I also took the chance to improve the naming in the codebase around said nodes, and to make the code a bid cleaner.

Fixes #293

### DIFF
--- a/include/proper_internal.hrl
+++ b/include/proper_internal.hrl
@@ -37,10 +37,6 @@
         {ok, Pid, Node} ->
             register(Node, Pid),
             _ = update_worker_node_ref({Node, {already_running, false}}),
-            Node;
-        {ok, Pid} ->
-            Node = list_to_atom(lists:concat([Name, "@", net_adm:localhost()])),
-            register(Node, Pid),
             Node;).
 -define(STOP_PEER_NODE(Name),
         peer:stop(Name)).
--- a/include/proper_internal.hrl
+++ b/include/proper_internal.hrl
@@ -27,6 +27,28 @@
 
 -include("proper_common.hrl").
 
+%%------------------------------------------------------------------------------
+%% Peer node module definition
+%%------------------------------------------------------------------------------
+
+%% OTP25 introduced `peer` to replace the `slave' module.
+%% It has a different API, so for now we have to live with macros to
+%% conditionally compile the proper code to handle the node setup.
+%% This whole setup can be removed once PropEr requires at the very least OTP25 to run.
+-if (?OTP_RELEASE >= 25).
+-define(START_PEER_NODE(Name),
+        peer:start(#{name => Name})).
+-define(STOP_PEER_NODE(Name),
+        peer:stop(Name)).
+-else.
+-define(START_PEER_NODE(Name),
+        begin
+            HostName = list_to_atom(net_adm:localhost()),
+            slave:start_link(HostName, Name)
+        end).
+-define(STOP_PEER_NODE(Name),
+        slave:stop(Node)).
+-endif.
 
 %%------------------------------------------------------------------------------
 %% Random generator selection
--- a/include/proper_internal.hrl
+++ b/include/proper_internal.hrl
@@ -31,21 +31,18 @@
 %% Peer node module definition
 %%------------------------------------------------------------------------------
 
-%% OTP25 introduced `peer` to replace the `slave' module.
-%% It has a different API, so for now we have to live with macros to
-%% conditionally compile the proper code to handle the node setup.
-%% This whole setup can be removed once PropEr requires at the very least OTP25 to run.
 -if (?OTP_RELEASE >= 25).
--define(START_PEER_NODE(Name),
-        peer:start(#{name => Name})).
+-define(CASE_START_PEER_NODE(Name),
+    case peer:start(#{name => Name}) of
+        {ok, Pid, Node} ->
+            register(Node, Pid),
+            _ = update_worker_node_ref({Node, {already_running, false}}),
+            Node;).
 -define(STOP_PEER_NODE(Name),
         peer:stop(Name)).
 -else.
--define(START_PEER_NODE(Name),
-        begin
-            HostName = list_to_atom(net_adm:localhost()),
-            slave:start_link(HostName, Name)
-        end).
+-define(CASE_START_PEER_NODE(Name),
+        case slave:start_link(_HostName = list_to_atom(net_adm:localhost()), Name) of).
 -define(STOP_PEER_NODE(Name),
         slave:stop(Node)).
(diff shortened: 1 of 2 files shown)


kostis:
Thanks for the PR @pablocostass !

Before this gets merged, I would like to request some changes to it though, which I will leave as comments in its code.
Once these are done, I see no reason not to merge this.

kostis (changes requested):
See comments.

kostis:
> I think Dialyzer is inferring in OTP25 from the register(Node, Pid) call that the type of Node is that of Pid, and therefore the old spec for update_worker_node_ref/1 won't match as node() =/= pid():

This does not make any sense.  Can you remove the two occurrences of `pid()` from the code so that I investigate?

pablocostass:
Yeah no, you are right it doesn't make any sense; I am in need of my second round of coffee, my bad.

The true reason why that error is there on OTP25, is because `?CASE_START_PEER_NODE(Name)` unfolds on a call to `peer:start/1` (which made me realize it should be a call to `start_link/1`), and the typing of that function is:
```erlang
-spec start_link(start_options()) -> {ok, pid()} | {ok, pid(), node()} | {error, Reason}
              when Reason :: term().
```

So it makes sense that Dialyzer is assuming that the pattern match of `{ok, Node} ->` in OTP25 has the type of `{ok, pid()}`

kostis:
> So it makes sense that Dialyzer is assuming that the pattern match of {ok, Node} -> in OTP25 has the type of {ok, pid()}


Once again, this sentence does not make sense.  Not because I do not understand Dialyzer's reasoning, but because I do not understand the reason for having confusing code in PropEr's code base.  If `peer:start_link/1` returns `{ok, Pid}` then we should have an `{ok, Pid} ->` case for it, not one named `{ok, Node} ->` just to save some lines.

More generally, I am really confused on what calls to `peer` should be use and what the code should do -- that's why I asked for help.  Note that, [according to peer's code](https://github.com/erlang/otp/blob/master/lib/stdlib/src/peer.erl#L190) both `peer:start_link/1` and `peer:start/1` return the same terms, namely `{ok, pid()} | {ok, pid(), node()} | {error, Reason}`.

I am assuming that when the documentation mentions `pid()` it means a Process ID and when it mentions `node()` it means an Erlang node.  So, I do not understand why in 25.0 code the `{ok, Pid, Node}` case should be:

```erlang
       {ok, Pid, Node} ->
           register(Node, Pid),
           _ = update_worker_node_ref({Node, {already_running, false}}),
           Node;
```

while for the `{ok, Pid}` return of `peer:start_link` (or `peer:start/1`) we should do:
```erlang
       {ok, Pid} ->
            _ = update_worker_node_ref({Pid, {already_running, false}}),
           Pid;
```

Something seems wrong there...

Note that all this code is in body of function whose spec is:
```erlang
-spec start_node(node()) -> node().
```
while this function does not return `node()` in the `{ok, Pid}` case, does it?

Do we have any tests that test both cases in 25.0 ??

pablocostass:
As you already know, the extension of PropEr with a parallel execution was done prior to OTP25 and the introduction of `peer`, in the times when the (terribly named) `slave` module was yet to be deprecated.

Back then, I wrote the code with the idea to start a node with `slave:start_link(Host, Name) -> {ok, Node}`, write down that `node()` name and stop it after the tests had been run with `slave:stop(Node) -> ok`.

`peer`, the module introduced in OTP25 to replace the aforementioned one, has a slightly different API and usage, and it yields as you pointed out different return values with both `peer:start/1` and `peer:start_link/{0,1}` (`{ok, pid()} | {ok, pid(), node()} | {error, Reason}`).

`peer` nodes are started with `peer:start_link(#{name => Name})`, which will give us both a `node()` name, and the `pid()` of the control process associated with it, and are stopped using the pid of the control process. The only cases on which starting a `peer` node will give us only the pid (`ok, pid()}`) is when the node has been started with the start option of `wait_boot` set to `false`, as the peer node will be started asynchronously. E.g.:

```erlang
1> _ = net_kernel:start([proper_leader, shortnames]).
{ok,<0.86.0>}
(proper_leader@PABLO-COSTAS-SANCHEZ)2> peer:start(#{name => proper_follower_1}).
{ok,<0.93.0>,'proper_follower_1@PABLO-COSTAS-SANCHEZ'}
(proper_leader@PABLO-COSTAS-SANCHEZ)3> peer:start(#{name => proper_follower_2, wait_boot => false}).
{ok,<0.98.0>}
```

In former of cases, to be able to reuse the code that is already written, we just need to register the node name with the pid of the control process and then the `peer` node can be stopped with `peer:stop(Name)`.

```erlang
{ok, Pid, Node} ->
    register(Node, Pid),
    _ = update_worker_node_ref({Node, {already_running, false}}),
    Node;
```

In the latter of the cases, which should not happen in a PropEr execution anyways, we create what would have been the node name ourselves, and register it with the pid of the control process, which allows us again to keep the code as it is.
```erlang
{ok, Pid} ->
    Node = list_to_atom(lists:concat([Name, "@", net_adm:localhost()])),
    _ = update_worker_node_ref({Pid, {already_running, false}}),
    Node;
```

And thus, the following spec holds true:
```erlang
-spec start_node(node()) -> node().
```

 
> Do we have any tests that test both cases in 25.0 ??
No we do not, given that unless a user of PropEr really goes out of their way to make the second case happen (i.e., starting the nodes with the proper names beforehand), it will not occur in a PropEr execution. And, besides, if someone were to do that, they would find themselves learning that whilst the `slave` module would return `{error, {already_running, Node}}` when trying to start nodes with a name already taken by a node, `peer` nodes will crash after a timeout.

kostis (approved):

PR #296, opened 2022-06-27
Introduce `?TYPE` macro

michalmuskala:
It allows explicitly using PropEr's "native types" feature.

This is more powerful than the current automatic parse transform-based inference:
* it can be used outside of `?FORALL` to produce PropEr types & combine
  them with other PropEr or native types.
* it allows using the "native types" feature with the proper parse transform disabled
* it allows expressing more types directly, without creating additional type aliases,
  for example: `?TYPE(atom() | float())`.

The semantics are identical to automatic inference, if the expression was wrapped
in a simple `id` type (when the syntax allows expressing such a type). For example:

```erlang
-record(foo, {}).
-type id(X) :: X.

% semantically equivalent definitions:
?FORALL(X, id(#foo{}), Prop).
?FORALL(X, ?TYPE(#foo{}), Prop).
```

Original PR message:
<details>
This allows running proper, with the `PROPER_NO_TRANS` flag, with no
loss of functionality in practice, by replacing the implicit type usage
introduced by the parse transform, by explicit calls to a new `?TYPE` macro.

In practice, this is even more powerful than the parse transform. In particular,
it's possible to define ad-hoc types with `?TYPE(X | Y)`, where before defining
a one-off type alias was necessary.
</details>

### DIFF
--- a/src/proper_typeserver.erl
+++ b/src/proper_typeserver.erl
@@ -166,7 +166,7 @@
 %%% <ul>
 %%% <li>It is allowed in any position to produce a PropEr type, including
 %%%   outside of `?FORALL'.</li>
-%%% <li>The same restriction on allowed recursive native types as in `?FORALL`
+%%% <li>The same restriction on allowed recursive native types as in `?FORALL'
 %%%   apply.</li>
 %%% <li>There is no risk of confusion between expressions and native types.
 %%%   Inside `?TYPE' everything is interpreted as a native type; for example,


kostis:
Thanks for the PR, but I have some questions / concerns about it.

I understand the rationale for allowing the possibility to run PropEr without the parse_transform, but the current description of the PR is very vague (and slightly contradictory).  Most notably, it does not come with any description of what is allowed (or not allowed) to be the argument of the `?TYPE` macro. Module local types only (as the single, very simple, example shows) or remote types too?  Can one specify records there and how are these handled?  In short, does it allow for everything that the current language of types allows or not?

Second, can I use `?TYPE` independently of whether the `PROPER_NO_TRANS` flag is set (when compiling the module)? 
- If no, why do I have to specify both (as in the simple example)?
- If yes, what does/should it do in this case?

Third, I am a bit uncertain whether allowing to write `X | Y` types are a good idea.  Note that this suddenly allows for the type notation to be used in arbitrary places in the code, when before this PR only Erlang terms were allowed.  Is this a good idea?

Finally, but perhaps these are only minor issues, I have a slight trouble with the phrasing that the PR message uses.

1. What does
    > with no loss of functionality in practice
   
   mean?  That there is some loss in theory?  If so, what?

2. I do not see anything that is _replaced_ here.  I only see the introduction of a new `?TYPE` macro that adds some particular functionality.  What exactly is _replaced_ ?  (or is it just unfortunate phrasing)

michalmuskala:
Hi @kostis, thank you for your feedback. I updated the description to make it clearer, and I've also updated the documentation to mention the `?TYPE` macro.

Indeed, this is adding a new feature. The fact that it enables using PropEr's "native types" feature without the parse transform is a side-effect (quite desirable to me, though, and the initial trigger for developing this feature).

I believe the updated commit message should address your feedback. I believe the only element that is not addressed is: "is this a good idea":

I believe it is, making PropEr more powerful and succinct in what it can express. Moreover, the `X | Y` are already partially supported by PropEr, just not in all the cases, for example:
```erlang
-type id(X) :: X.

?FORALL(X, id([atom() | float()]), ...).
```
I believe this change makes PropEr more consistent allowing the same syntax to be used everywhere and not be subject to the limitations of the Erlang parser/compiler.

src/proper_typeserver.erl
```diff
@@ -161,17 +161,19 @@
...
 %%% <li>It is allowed in any position to produce a PropEr type, including
-%%%   outside of `?FORALL`.</li>
-%%% <li>The same restriction on allowed recursive native types as in `?FORALL` apply.</li>
-%%% <li>There's no risk of confusion between expressions and native types,
-%%%   inside `?TYPE` everything is interpreted as a native type, and as such
-%%%   `?TYPE([integer()])` will produce an arbitrary integer list.</li>
-%%% <li>It is not checked for correct syntax - using invalid syntax can produce
-%%%   errors when the property is evaluated, which are hard to understand.</li>
+%%%   outside of `?FORALL'.</li>
+%%% <li>The same restriction on allowed recursive native types as in `?FORALL`
```
pablocostass, line 169: missed this one:
```suggestion
%%% <li>The same restriction on allowed recursive native types as in `?FORALL'
```

michalmuskala:
Is there anything missing here from my side?

PR #299, opened 2022-09-22
Add 25.1.x in tested OTP versions

kostis:
While at it, do some minor copyright-related updates.


kostis:
There is a regression in EUnit shipped with Erlang 25.1.
This regression causes `rebar3 eunit` to fail when it loads test suites.
See erlang/otp#6320.

PR #302, opened 2022-12-26
readme: fix correct git clone URI

kikofernandez:
This commit fixes the URI for the call `git clone`. It used to be:
```git
git clone git://github.com/proper-testing/proper.git
```
but it has been replaced by:
```git
git clone https://github.com/proper-testing/proper.git
```
I got timeouts when calling `git clone git://github.com/proper-testing/proper.git`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ Quickstart guide
     base:
 
     ```shell
-    git clone git://github.com/proper-testing/proper.git
+    git clone https://github.com/proper-testing/proper.git
     ```
 *   Compile PropEr: Simply run `make` if you just want to build PropEr.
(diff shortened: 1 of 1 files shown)


pablocostass:
Good catch! It should have either been either the HTTPS URI (your fix) or the SSH URI (`git@github.com:proper-testing/proper.git`), but it seems both options got mixed up resulting in an improper URI

kostis:
Thanks for your interest in PropEr and thanks for noticing this.  (Also, season's greetings to both of you!)

As mentioned in #303, which extends testing with the recently released Erlang/OTP 25.2, since there would be changes in the README file anyway, I folded the fix there (using the SSH URI).  Hope this is OK.  I will close this PR.

Issue #290, opened 2022-05-24
Atom generator exhausts the atom limit


Maria-12648430:
The `atom()` generator produces atoms at random, with no limitations whatsoever. By extension, this is also true for union generators like `term()`. As atoms are never garbage collected, this means that over time the limit on the number of atoms in a runtime instance will get exhausted. How quickly that happens depends on how generators are used, how many properties there are in a suite, and how many suites there are.

We are working on property tests for OTP, currently for the `lists` module. We have 71 properties so far and are not done yet, but we already encounter failures because the atom space in exhausted. Yes, we can tune our generators to generate atoms more economically (cumbersome and makes tests more artificial), but still, this puts a tight lid on things in general. Going to extremes, it would never be possible to equip all of OTP with property tests, not even all of `stdlib`.

kostis:
Atom garbage collection is something known to be missing from Erlang/OTP's implementation for many years now, and a limitation which has forced most Erlang developers to write ugly and/or unnatural code in many of their programs.  Moreover, the implementation of an atom garbage collector is not rocket science. So, perhaps it's high time that it is implemented.

If property-based testing is something that is of use to Erlang/OTP and your particular way of using the `atom()`/`term()` generator causes trouble for OTP, perhaps this may give some extra incentive to the OTP folks to bite the bullet and finally implement atom garbage collection.

In my opinion, artificially limiting the `atom()` and `term()` generators to e.g. just the existing atoms is not the PropEr solution to the issue(s) you are facing.  But I am willing to be convinced otherwise.

Maria-12648430:
Well, I have no real idea how that works internally, I have my guesses and some cloudy memory. But I think it's not just a simple matter of not wanting to bother of cleaning up after yourself :stuck_out_tongue_winking_eye: And I don't think my particular way of using `atom()` and `term()` will induce the OTP team to go down that road, despite them being in favor of property tests :wink:

That said, personally I didn't feel any need for more or infinite atoms, and I would second-guess any production code that needs it.

On the other hand, I understand that limiting the generators is contrary to what property testing is really about: generating all sorts of possible input. But I don't see how the two will be brought together :woman_shrugging:

Maria-12648430:
FWIW: https://erlangforums.com/t/on-garbage-collecting-atoms/1476

Maria-12648430:
[...]

PR #307, opened 2023-03-21
Add ability to limit the number of generated atoms

Maria-12648430:
As mentioned in #290, the usage of the atom generator (explicit or implicit) can lead to exhaustion of the atom table. While I understand that PropEr does not want to restrict itself to just afew atoms per se, it would be very useful if a user (like me) was given the choice to limit this if needed.


src/proper_gen.erl
```diff
@@ -406,12 +406,46 @@ float_gen(_Size, Low, High) ->
...
+        ?LET(Atom,
+             try persistent_term:get({'$proper', ?MODULE, ?FUNCTION_NAME, Size}) of
+                 {N, Atoms} when N =< Limit ->
+                     proper_types:oneof(Atoms);
+                 {N, CurAtoms} ->
+                     ?LET(MoreAtoms,
+                          proper_types:vector(Limit-N, atom_gen1(Size)),
+                          begin
+                              Atoms = CurAtoms ++ MoreAtoms,
+                              persisten_term:put({'$proper', ?MODULE, ?FUNCTION_NAME, Size},
```
kostis, line 424: This line cannot possibly have ever executed...

src/proper_gen.erl
```diff
@@ -406,12 +406,46 @@ float_gen(_Size, Low, High) ->
...
+                                                  {Limit, Atoms}),
+                              proper_types:oneof(Atoms)
+                          end
+                     )
+             end,
+        Atom)).
+
+atom_gen1(Size) ->
     ?LET(Str,
-	 ?SUCHTHAT(X,
```
kostis, line 410: What _exactly_ is the purpose of the changes to the following lines?

src/proper_gen.erl
Maria-12648430: Yeah, just noticed myself :sweat_smile:

I probably should have said that this is not yet finished, just wanted to show it to see if the general idea will be accepted. Therefore, there are no tests yet.

src/proper_gen.erl
Maria-12648430: No purpose, that happened because of auto-formatting. I can change that back later if you want.

src/proper_gen.erl
Maria-12648430: Typo corrected.

src/proper_gen.erl
Maria-12648430: Formatting corrected, also in other places.

kostis (changes requested):
Not sure there is something to review at this point: this PR comes with no tests for specifying the `atom_limit` option, even with `infinity`, so I have zero confidence that the code does anything reasonable when that option is set.  

When you do add some tests, you will discover that the code for the non-infinity case cannot run on older Erlang/OTP versions (e.g. 21 and 22 - I believe) because `persistent_term` is not available there.  So, this code cannot be included as is as long as PropEr still supports these Erlang/OTP versions...

Maria-12648430:
> When you do add some tests, you will discover that the code for the non-infinity case cannot run on older Erlang/OTP versions (e.g. 21 and 22 - I believe) because `persistent_term` is not available there. So, this code cannot be included as is as long as PropEr still supports these Erlang/OTP versions...

For that case, I suggest we put in a conditional that falls back to always using `infinity` when used with an older OTP version that does not support `persistent_term`. Would that be acceptable?

kostis:
> EDIT: persistent_term was introduced in OTP 21.2.

OK - so then it's not that bad as I feared.
In that case, yes, the conditional you suggest will do.

PR #319, opened 2025-01-21
Add generators for UTF-8 strings

### LINKED ISSUE #318: proper_types:char() and proper_types:string() may generate invalid Unicode
I had a proper test fail, because `proper_types:string()` returned `[55296]` which is `[16#D800]`, and that's not a valid Unicode string, causing `unicode:characters_to_list/1` (and others) to fail on it.

I believe the root cause is that `proper_types:char()` returns any integer between `0` and `16#10ffff`, without taking care to avoid invalid Unicode code points.

kostis:
As reported in #318, `proper_types:string()` may generate invalid Unicode strings; the Erlang type language overapproximates the set of character lists that represent valid Unicode strings.

This PR extends the `proper_unicode` module with functions that generate valid UTF-8 character lists.

Like most `proper_types` generators, these functions are auto-imported and thus one can write
```erlang
   ?FORALL(Str, utf8_string(), ...)
```
without having to specify the `proper_unicode:` module prefix.

Closes #318

### DIFF
--- a/include/proper.hrl
+++ b/include/proper.hrl
@@ -1,6 +1,6 @@
 %%% -*- coding: utf-8; erlang-indent-level: 2 -*-
 %%% -------------------------------------------------------------------
-%%% Copyright 2010-2022 Manolis Papadakis <manopapad@gmail.com>,
+%%% Copyright 2010-2025 Manolis Papadakis <manopapad@gmail.com>,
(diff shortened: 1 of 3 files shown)


kostis:
There has not been any feedback, so I will interpret this as an OK for this PR.  Merging.

PR #322, opened 2025-04-02
Bump to 1.5.0 for hex publishing

ferd:


kostis:
I seem to have missed this file in the release.  Thanks for catching it.  However, the release version is 1.5.0, so I've updated the file accordingly.  Once the tests pass, I will merge it.

PR #324, opened 2026-02-18
enable building proper with OTP-29

mikpe:
- eliminate old-style catches
- resolve is_record/1 ambiguity
- fix variable exported from subexpression

### DIFF
--- a/test/proper_exported_types_test.erl
+++ b/test/proper_exported_types_test.erl
@@ -52,7 +52,13 @@ not_handled() ->
   R = [{M,T,A,proper_typeserver:demo_translate_type(M, stringify(T, A))}
        || {M,T,A} <- MTs],
   {OKs,Errors} = lists:partition(fun type_translation_is_ok/1, R),
-  {[Inst || TGen <- OKs, (Inst = pick_instance(TGen)) =/= ok], length(Errors)}.
+  {lists:filtermap(fun filtermap_instance/1, OKs), length(Errors)}.
+
+filtermap_instance(TGen) ->
+  case pick_instance(TGen) of
+    ok -> false;
+    Inst -> {true, Inst}
+  end.
 
 pick_instance({M,T,A,{ok,Gen}}) ->
   {ok,Inst} = proper_gen:pick(Gen),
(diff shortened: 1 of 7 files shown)


test/proper_exported_types_test.erl
```diff
@@ -52,7 +52,13 @@ not_handled() ->
   R = [{M,T,A,proper_typeserver:demo_translate_type(M, stringify(T, A))}
        || {M,T,A} <- MTs],
   {OKs,Errors} = lists:partition(fun type_translation_is_ok/1, R),
-  {[Inst || TGen <- OKs, (Inst = pick_instance(TGen)) =/= ok], length(Errors)}.
```
kostis, line 55: I want to understand the rationale for this change and what _exactly_ is wrong with this code in general and in the upcoming OTP 29.  To me, this code besides being "valid Erlang", is perfectly understandable in what it does and how it achieves it.  Moreover, it's the last statement in the body of this function, so there is not even a chance that variables will be used after it.  So why is it better to have to write a new function of six lines for it?

Also, I've noticed that the [release notes for the pre-releases of 29](https://github.com/erlang/otp/releases/tag/OTP-29.0-rc2) advertise a "new language feature" that _encourages_ such code.  I quote from that page:

> By enabling the compr_assign feature, it is now possible to bind variables in a comprehensions. For example: `[H || E <- List, H = erlang:phash2(E), H rem 10 =:= 0]`

How does the PropEr code differ from this example and what am I missing?

kostis (commented):
Thanks for the PR.  I've left a comment concerning one of the changes.

test/proper_exported_types_test.erl
mikpe: It's due to an unfortunate combination of two language changes:
1. Exporting bindings from subexpressions is deprecated. In this case the lhs of `(Inst = ...) =/= ok` triggers that. Plain matches don't count as subexpressions.
2. Previously you could write `[Y || X <- L, Y = f(X), predicate(Y)]` and it would compile but give a `bad filter` exception at runtime. The other change, enabled by `compr_assign`, makes those bindings work.

I've experimented a bit and found a shorter workaround: instead of having a match (`Y = f(X)`) to bind the intermediate value, use a single-element generator (`Y <- [f(X)]`). I'm pushing an update with that change.

kostis:
Thanks again for your PR. Now that 29.0 has been released, I've extended the CI to also include it and merged this.

PR #327, opened 2026-03-20
Prepare for OTP29

NelsonVides:
Mostly issues around deprecating `catch` expressions. Also a warning around the is_record bif and deprecated export from subexpression warning (the new filtermap).

### DIFF
--- a/src/proper_erlang_abstract_code.erl
+++ b/src/proper_erlang_abstract_code.erl
@@ -67,6 +67,8 @@
 %%% '''
 -module(proper_erlang_abstract_code).
 
+-compile({no_auto_import, [is_record/1]}).
+
 -export([module/0, module/1, guard/0, guard/1, expr/0, expr/1]).
 
 -export([term/0, term/1]).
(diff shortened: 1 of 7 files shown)


NelsonVides:
Actually I just noticed that #324 does pretty much the same and it's actually kinda cleaner, I like that one better 😄
Looking forward for a new OTP29-ready release 🥹

There's also the option to add `latest` to the CI matrix, in case we'd want to verify this in CI as well.

kostis:
Thanks for your PR and your comment.  I agree with you that #324 is a cleaner PR, so we'll probably go with that.

> There's also the option to add latest to the CI matrix, in case we'd want to verify this in CI as well.

Yes, we should definitely verify this in the CI - that's the point in having one.  However, AFAIK Erlang/OTP 29 is not yet released.  Does `latest` also consider pre-releases as latest candidates?

NelsonVides:
Yes, `latest` includes release candidates: https://github.com/erlef/setup-beam?tab=readme-ov-file#latest-versions 🙂 

I can't push to the other PR, but maybe we want to add that configuration to the CI matrix and get the other one merged? I'm closing this in the meantime in favour of #324 💪🏽

